PR #48, opened 2021-03-20
Configures the github CI to run static code check on Saskatoon code

tristanlatr:
It does not fixes the issues raised by `pyflakes` and `mypy`, I'll get to that later. 

Also I needed to switch to `setup.py` to manage dependencies, more details on the commit message. 

Fixes #27

PR #49, opened 2021-03-20
List model imports and fix bug

tristanlatr:
To fix pyflakes issues and to have a more clear Python code for newcomers: this PR lists all classes imports instead of using `from .. import *`. when importing models.

I've also commented a few unused lines of code too, I figured they might be useful in the future. 

Marked as draft until #48 is merged. 

This PR also solves a bug in the publishing a pick logic (found with static checks).


tiagovaz:
Hi Tristan, I've just merged the last PR. Please have a look on this one and let me know when it's ready. Also, notice that in the meantime I updated Django version to >=3.0.14 in setup.py due to a recent CVE (so one more reason to git pull right before the PR).

Also, don't bother with unused imports and other code cleaning for now. I'll have a better time for code cleaning, giving that I'm still migrating stuff from the old version and it's normal to have half-functionalities added. 

Last thing, it's easier at this point to have less changes in each PR, lots of work going on at the same time (you, me and tassia!), so big chunks requires lots of time for reviewing.

Thanks,

tristanlatr:
Hi Tiago, 

I'm done for this PR :)

The pytype checks now passes ! I will now gradually add type annotation in the code base. 

I think I've fixed a bug in b940615.

Thanks

PR #50, opened 2021-03-27
Revert "Use a .env to store Saskatoon configuration "

tiagovaz:
Reverts LesFruitsDefendus/saskatoon-ng#45

Broken, needs further review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,3 @@
-# DS Store files
-.DS_Store
-
-# Settings file
-.env
-
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -50,37 +50,25 @@ To install Python requirements use :
 pip3 install -r requirements.txt
 ```
 
-## Database and other settings
+## Database
 
-To set new settings, adapt the following ``.env`` 
-file and place it inside `saskatoon/` project directory. 
+Edit the DB path file at `saskatoon/saskatoon/settings.py`, section `DATABASES`:
 
 ```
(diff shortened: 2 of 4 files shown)


tassia:
In fact it works, just missing the .env file. Better to maintain that file in the repo with default values, to be changed if desired.

@tiagovaz: safe to revert the revert, just add the following content in a .env file:

# SECURITY WARNING: keep the secret key used in production secret!
# More infos: https://docs.djangoproject.com/fr/3.1/ref/settings/#secret-key
SASKATOON_SECRET_KEY='%5k4)igg&h!k(@3o!5x%6_)y6go#sfbf5opqf-la3h05&5hhp+'

# SECURITY WARNING: don't run with debug turned on in production!
SASKATOON_DEBUG=no
# Database settings
SASKATOON_DB_ENGINE=django.db.backends.mysql
SASKATOON_DB_NAME=saskatoon_prod
SASKATOON_DB_USER=saskatoon
SASKATOON_DB_PASSWORD=
SASKATOON_DB_HOST=127.0.0.1
# Misc
SASKATOON_TIME_ZONE=UTC

tristanlatr:
The secret key should not be public I think right ?

Also i've added the .env to the list of gitignore not commit my config. Maybe we can include a default.env file instead ? Then wed have to rename it to actally use it with saskatoon. That shouls be documented in the INSTALL.md.

tiagovaz:
This key is not in production and can be kept for local instances. Once in production, a new key will be generated, then we can ignore .env from git.

tassia:
Tristan, can you re-issue the PR? Github is having trouble to revert the revert. Thanks!

PR #70, opened 2021-05-04
Setup tests

tristanlatr:
This PR adds the basic setup to automate the testing of Saskatoon with selenium.

I've add only some dummy tests for now. 

It does not run into the github CI though, only locally once setup with a test superuser.


tristanlatr:
I'm working on another PR that really adds some real use cases tests.
I'll submit it for review once this PR has been approved and merged

tristanlatr:
I don't get it, I't working OK locally...

tristanlatr:
I think we can disregard the failure for now as it's only failling in the CI.

qusvv:
Hi Tristan! I'm looking at your setup-tests branch and there are a couple changes I would like to make as I set it up. What's the best way to proceed, should I push to origin/setup-tests? An other way would be to integrate your changes to the develop branch and add my commits to it. Let me know :)

tristanlatr:
I'll let you merge this PT in `develop` branch ! :D 

Thanks

qusvv:
oops and I renamed SASKATOON_ADMIN_EMAIL and SASKATOON_ADMIN_PASSWORD (see last commit)

tristanlatr:
Hi @qusvv,

Looking at the code I think there is a bug in the way you create the `WebDriver` instance: It should be created inside the following block, not at the module level:
```python
@pytest.fixture(scope="module")
def driver():
    #here
```

This was the heart of the issue. Because the `driver.quit()` method is called when a test has been executed, this makes the driver unusable again.

The issue is now fixed! I've pushed some commits to the `develop` branch. I've also removed some unnecessary things from the tests setup .

And I'll see if I can make is work with Firefox, as it's also a popular browser. 

Also, thanks for the documentation about the web driver and everything.

I'll close this PR since you already merged it manually to `develop` branch. Maybe next time we can use the Github PR / review system ? 

Thanks.

qusvv:
@tristanlatr  This looks great! Sorry I shouldn't have messed up with your code... I totally got confused by the function annotations, thinking you were passing webdriver.Chrome 'as' driver... Ugh... 

And yes sorry for the weird workflow... I should have tried Chrome right away instead of fiddling with your PR, I'll use the review system next time.

So runtests.sh now works for me on Chrome. I run into issues with `tox -e test` though:
```
E selenium.common.exceptions.WebDriverException: Message: unknown error: Chrome failed to start: exited abnormally.
E (unknown error: DevToolsActivePort file doesn't exist)
E (The process started from chrome location /usr/lib/chromium/chromium is no longer running, so ChromeDriver is assuming that Chrome has crashed.)
```
Do you also experience this? I'm not familiar with tox. It's only supposed to run `runtests.sh` as defined under `[testenv:test]` in `tox.ini` right?


qynn

tristanlatr:
Hi Qynn,

Great that we figured out :)

Yes the `tox -e test` is only supposed to run the the run command inside a virtual environment created for the occasion. 

So I'm not sure what is the cause of this error...

PR #80, opened 2021-05-25
Apidocs

tristanlatr:
Add a simple script to generate the API documentation of the Python objects every time we push. 

It goes to: https://lesfruitsdefendus.github.io/saskatoon-ng/


qusvv:
@tristanlatr  Sweet!  `tox -e apidocs` ran fine on my machine

Do you have any specific template for doctrings you like to follow? I am getting used to comment functions but I rarely add doctrings on the top of modules

tristanlatr:
@qusvv,

No I don't have a specific template. But pydoctor parses the docstrings with a pretty strict reStructuredText parser. You can read the docs here: https://pydoctor.readthedocs.io/en/latest/docformat/restructuredtext.html

Regarding how to document you can read this: https://pydoctor.readthedocs.io/en/latest/codedoc.html

Note that the search feature doesn't work if your viewing the files on your local system, this is why we have it there:  https://lesfruitsdefendus.github.io/saskatoon-ng/ . You could also use ``python3 -m http.server --dir ./apidocs`` to view it locally

PR #92, opened 2021-06-04
Fixtures: added comment + participation request

qusvv:
Added 1 comment + 1 participation_request fixtures. I had issues with the send_email function: ConnectionRefusedError: [Errno 111] Connection refused). So in order for `saskatoon/fixtures/init` to pass I needed to set fail_silently=True in `send_email()` from `harvest/signals.py.`

### DIFF
--- a/saskatoon/fixtures/member-authuser.json
+++ b/saskatoon/fixtures/member-authuser.json
@@ -10,7 +10,7 @@
     "email": "pick@leader.com",
     "date_joined": "2021-05-26T21:26:41.699Z",
     "is_active": true,
-    "is_staff": false,
+    "is_staff": true,
     "groups": [],
     "user_permissions": []
   }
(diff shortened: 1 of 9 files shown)


saskatoon/fixtures/member-authuser.json
```diff
@@ -10,7 +10,7 @@
     "email": "pick@leader.com",
     "date_joined": "2021-05-26T21:26:41.699Z",
     "is_active": true,
-    "is_staff": false,
+    "is_staff": true,
```
qusvv, line 13: in `harvest/api.py`
```
class PickLeaderAutocomplete(autocomplete.Select2QuerySetView):
    def get_queryset(self):
       .....
        qs = AuthUser.objects.filter(is_staff=True)
       .....
```

Issue #19, opened 2020-05-17
Improve the fruits recipient data entries process


tristanlatr:
See: https://github.com/tiagovaz/saskatoon/issues/87

qusvv:
1. Filter Trees based on `harvest.trees`
2. Preload inline forms for each recipient
3. Fix total of lbs (multiplied by 10 in French)

PR #96, opened 2021-06-06
Improve Fruit Distribution + RFPManageForm

qusvv:
Issue #19 : improve fruit distribution form:

- added a collapsible form to add a new recipient
- only owner + pickers + pick-leader + beneficiary organizations are listed in the recipient dropdown
- only trees selected for the harvest are listed in the tree dropdown
- added a delete button with confirmation

The only drawback that i see is that when adding or deleting a distribution page reloads on top and we still need to scroll back down but I think it is a way less tedious process than before.

Warning: along the way I have added the `django-debug-toolbar` which i find very useful to check which static files have been loaded, among other things: to install it: `$(venv) pip3 install django-debug-toolbar`


saskatoon/saskatoon/settings.py
```diff
@@ -77,6 +80,7 @@
     'rest_framework',
     'django_filters',
     'crispy_forms',
+    'debug_toolbar',
```
tristanlatr, line 83: Maybe the debug toolbar could be enabled only if `SASKATOON_DEBUG` is True ?

saskatoon/sitebase/static/js/confirm/delete.js
```diff
@@ -0,0 +1,8 @@
+
+function confirmDelete() {
+    if (confirm('Are you sure you want to delete this entry?')) {
+        // console.log("Entry Deleted");
```
tristanlatr, line 4: Should it return True ?

tristanlatr (approved):
It looks great! 
I haven't tested it yet though.

saskatoon/sitebase/static/js/confirm/delete.js
qusvv: No when adding `return true;` the entry will get deleted when you click cancel. I think you just want it to pass. 
I also tried:
```
function confirmDelete() {
      return confirm('Are you sure you want to delete this entry?')
}
```
but didn't work

saskatoon/saskatoon/settings.py
qusvv: I'm pretty certain django does that by default

tristanlatr: ok1

saskatoon/sitebase/static/js/confirm/delete.js
tristanlatr: Ok!

PR #122, opened 2022-03-18
Adjust icons

tristanlatr:
Use notika-checked and notika-close instead of arrow up and down to signify a feature exist. Make notika-checked green.


2ynn (approved):

tristanlatr:
Before:
by default 2022-03-29 at 10 53 17 AM

After:

by default 2022-03-29 at 10 52 20 AM

Issue #121, opened 2022-03-16
Install Rosetta to manage transations in human firendly way


tristanlatr:
See https://medium.com/@edadenizcelik/make-your-django-project-global-76095d280975 and https://django-rosetta.readthedocs.io/installation.html for more informations.

PR #114, opened 2021-09-20
Forms - list views - filters - user roles - and more

2ynn:
I pushed the develop branch to https://saskatoon-dev.lesfruitsdefendus.org

I would appreciate if you could test it, especially all the different forms:
- [ ] new owner based on pending_property (see the "register" button next to "contact" title in a pending property detail page)
- [ ] new person (pickleader, core-member, volunteer, etc..) 
- [ ] new (organization + contact person)
- [ ] new (property + owner)

I the meantime I will work on a document to list all the new features to test so the core members can try it too.

PR #128, opened 2022-03-26
install Rosetta to manage transations

victorphoenix3:
Install Rosetta to manage translations in human friendly way

fixes:#121

Screenshots: 

rosetta
manage translation

### DIFF
--- a/saskatoon/saskatoon/urls.py
+++ b/saskatoon/saskatoon/urls.py
@@ -12,4 +12,6 @@ urlpatterns = [
     path('accounts/', include('django.contrib.auth.urls')),
     path('i18n/', include('django.conf.urls.i18n')),
     path('__debug__/', include(debug_toolbar.urls)),
+    path('rosetta/', include('rosetta.urls'))
 ]
+
--- a/saskatoon/saskatoon/settings.py
+++ b/saskatoon/saskatoon/settings.py
@@ -74,6 +74,7 @@ INSTALLED_APPS = [
     'crispy_forms',
     'debug_toolbar',
     'django_extensions',
+    'rosetta'
 ]
 
 MIDDLEWARE = [
--- a/saskatoon/sitebase/templates/app/base/header.html
+++ b/saskatoon/sitebase/templates/app/base/header.html
@@ -50,6 +50,7 @@
                                                 {% translate "Manage beneficiary" %}</a></li>
                                         {% endif %}
                                         <li><a href="/admin/">{% translate "Admin panel" %}</a></li>
+                                        <li><a href="/rosetta/">{% translate "Manage translations" %}</a></li>
                                     </ul>
                                 </li>
                             {% endif %}
--- a/setup.py
+++ b/setup.py
@@ -42,6 +42,7 @@ setup(
         "django-redis",
         "django-dotenv",
         "django-crispy-forms",
+        "django-rosetta==0.9.8"
     ],
     extras_require      =   {
         'test': ['pytest', 'selenium', 'pytest-django', 'invoke', 'tox']


saskatoon/saskatoon/urls.py
```diff
@@ -1,5 +1,5 @@
 from django.contrib import admin
-from django.urls import path, include
+from django.urls import path, include, re_path
```
tristanlatr, line 2: I don't think we should use both `re_path` and `path` if `path` can handled all our cases. I'm not against `re_path`, it's just that if it's not necessary, we should stick with what we already use, in my opinion.

saskatoon/saskatoon/urls.py
```diff
@@ -10,3 +10,8 @@
     path('',include('django.contrib.auth.urls')),
     path('__debug__/', include(debug_toolbar.urls)),
 ]
+
+if 'rosetta' in settings.INSTALLED_APPS:
+    urlpatterns += [
+        re_path(r'^rosetta/', include('rosetta.urls'))
+    ]
```
tristanlatr, line 17: Since rosetta is always enabled, I think it's safe to do like other `urlpatterns`, and include it directly in the list.

```suggestion
    path('rosetta/', include('rosetta.urls')),
]
```

tristanlatr (changes requested):
Hello @victorphoenix3,

Thanks for your contribution! This is a very useful addition, and will facilitate the collective work. 

Have you already used rosetta ? (I haven't, but it looks pretty easy). 

I I understand correctly, rosetta is compatible with the translation files we already have in place, and we just need to add translations from the rosetta interface ?

Other question: what are the access level required to access rosetta? I suppose we have to be admin, but just want to make sure.

Another little thing that would be needed to accept this PR is, alongside the link to the admin panel, a "Manage translations" link.

by default 2022-03-26 at 11 47 44 AM
 
Talk to you later,

Tristan

tristanlatr:
I've changed the target branch to be `develop`, this should be used until we merge PR #114.
Sorry for the conflicts.

saskatoon/saskatoon/urls.py
victorphoenix3: done

saskatoon/saskatoon/urls.py
victorphoenix3: done

victorphoenix3:
Hi @tristanlatr 

> Have you already used rosetta ? (I haven't, but it looks pretty easy).

Ditto!! I did go through their documentation and It does look easy.
<br>
> what are the access level required to access rosetta

We need admin level access. I tried with non-admin roles but it redirects to login page again.
<br>
> rosetta is compatible with the translation files we already have in place?

Yes, it is mentioned in their documentation that they are compatible with`po` files
<br>


I have added a "Manage translations" link and attached a screenshot.

tristanlatr:
Hi @victorphoenix3, 

Thanks for your investigation.

The diff looks great, I’ll pull your branch in the next couple days and try it out. 

I have still a question: is editing the translation edits the the .po files directly ? It would be good to have to very short process to update translations in the documentation. Do you think you could add a few lines to make it clear ? 

Thanks a lot,

I look forward to work in translations now!

Tristan

2ynn:
@victorphoenix3 have you tried compiling translation files with `makemessages` ?
https://docs.djangoproject.com/en/4.0/ref/django-admin/#makemessages

tristanlatr:
Hi @victorphoenix3,

I've pull your branch locally and it works great! It's very easy to update translations. And yes editing the translation edits the the .po files directly, so it's very strait forward to update the translations locally and then submit a PR with update translation files. 

Thanks again, I think we can merge this PR.

tristanlatr (approved):

2ynn:
@victorphoenix3 @tristanlatr rosetta does not generate the updated .po files for us right? Can you look at this?

tristanlatr:
> rosetta does not generate the updated .po files for us right? Can you look at this?

I believe our .po files already exists. And yes rosetta edits the .po file, making it easy to submit a PR after having worked on translations.

2ynn:
qynn



Mar 29, 2022, 10:43 AM by ***@***.***:

tristanlatr:
Hi,

Yes I think we should update the .op files from scratch with makemessages. 
I'de suggest we open a new issue to do this and we merge this PR now.
We just need not to work on translations before to rebuild the .po files, I suppose. 

Thanks @victorphoenix3,

See https://github.com/LesFruitsDefendus/saskatoon-ng/issues/158 for the makemessages issue. @victorphoenix3 , do you want to work in this as well ? 

Talk to you later,

victorphoenix3:
@2ynn I think we need to run `makemessages` ourselves. I found this on StackOverflow. 

`you have to run python manage.py makemessages -a to update the catalog from your templates and python source after you've modified the templates`

PR #135, opened 2022-03-26
Add "<4.0.0" for Django required version

mina-andajos:
After using `pip3 install .` while following `INSTALL.md`, Django 4.0.3 got installed for some reason.
After running `python3 saskatoon/manage.py migrate --skip-checks`, I got an import error most probably because of version incompatibility.
So adding this restriction will prevent the behavior that occurred to me.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ setup(
         "asgiref>=3.2.7",
         "cffi==1.14.0",
         "cryptography>=3.3.2",
-        "Django>=3.0.14",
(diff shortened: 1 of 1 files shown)


mina-andajos:
[Link to a related stackoverflow issue](https://stackoverflow.com/questions/70524028/importerror-cannot-import-name-force-text-from-django-utils-encoding-usr)

tristanlatr:
Looks good, thanks!

PR #137, opened 2022-03-26
Fix load_dotenv() import

mina-andajos:
Replaced `read_dotenv()` with `load_dotenv()` as it isn't defined.

### DIFF
--- a/saskatoon/saskatoon/settings.py
+++ b/saskatoon/saskatoon/settings.py
@@ -15,7 +15,7 @@ https://docs.djangoproject.com/en/3.0/ref/settings/
 
 import os
 from pathlib import Path
-from dotenv import read_dotenv
+from dotenv import load_dotenv
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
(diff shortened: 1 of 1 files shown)


farhanmasud:
I think code change might not be needed for this. The develop branch uses [django-dotenv](https://github.com/jpadilla/django-dotenv) instead of [python-dotenv](https://github.com/theskumar/python-dotenv).

A fix for this issue is mentioned in the [django-dotenv documentation](https://github.com/jpadilla/django-dotenv#common-problems).

tristanlatr:
Thanks for the pull request and the analysis. 

The solution for the issue would be to drop python-dotenv requirement?
Or should we drop [django-dotenv](https://github.com/jpadilla/django-dotenv)?
@2ynn, do you have an opinion on this issue?

Thanks all

2ynn:
@tristanlatr @mina-andajos I had to switch to `django-dotenv` in develop branch because I had issues with the debian dev server. Note that `python-dotenv` is no longer in the requirements (setup.py) on the develop branch, so `read_dotenv` should not raise any issue (i didn't have any problem deploying on my side). @mina-andajos maybe your virtual environment still has `python-dotenv` installed (e.g. when installing requirements from the master branch) and that's why you get an error?

mina-andajos:
> @tristanlatr @mina-andajos I had to switch to `django-dotenv` in develop branch because I was having issues deploying on the debian dev server. Note that `python-dotenv` is no longer in the requirements (setup.py) on the develop branch, so `read_dotenv` should not raise any issue (i don't have any locally or on dev server). @mina-andajos maybe your virtual environment still has `python-dotenv` installed (e.g. when installing requirements from the master branch) and that's why you get an error?

that's what was causing the error yes. 👍

mina-andajos:
I should close the PR right?

tristanlatr:
Thanks @2ynn and everyone, 

Yes you can close this pull request.

Talk to you later,

Issue #124, opened 2022-03-24
Make the register button more evident 


tristanlatr:
The register button for newly created members of the [community](https://saskatoon-dev.lesfruitsdefendus.org/community/) is tiny.
I'de suggest something more obvious like "Register user" or "Reset password" if the user password's is already set.
We can keep the icon, but I feel like some text should be also present. Note that "register" text appears only on hover.

<img width="79" alt="by default 2022-03-24 at 12 37 14 PM" src="https://user-images.githubusercontent.com/19967168/159965961-4b9a002b-9984-4996-b463-43013ad264d7.png">

yuktea:
I would like to work on this.

victorphoenix3:
> I would like to work on this.

Hi @yuktea, I had already been working on this and didn't see your comment before submitting the PR.

victorphoenix3:
> I would like to work on this.

Hi @yuktea, I had already been working on this and didn't see your comment before submitting the PR.

PR #139, opened 2022-03-26
add buttons for register user and reset password

victorphoenix3:
fixes:  #124

### DIFF
--- a/saskatoon/sitebase/templates/app/list_views/community/data_row.html
+++ b/saskatoon/sitebase/templates/app/list_views/community/data_row.html
@@ -27,11 +27,12 @@
             <a href="/admin/member/authuser/{{user.id}}/password/"
                onclick="return confirm('{% trans "Are you sure you want to reset this user password?" %}');"
             {% if not user.password %}
-                title="register">
-                <small><i class="fa fa-plus"></i> </a></small>
+                title="register-user"
+                >
+                <small class="btn btn-primary  waves-effect btn-sm"><i class="fa fa-plus"> </i> Register User </a></small>
             {% else %}
                 title="reset-password">
-                <small><i class="fa fa-refresh"></i> </a></small>
+                <small><i class="fa fa-refresh"></i></a></small>
             {% endif %}
         {% endif %}
         <br>{{ user.person.phone|default:"" }}
--- a/saskatoon/sitebase/templates/app/list_views/community/view.html
+++ b/saskatoon/sitebase/templates/app/list_views/community/view.html
@@ -7,8 +7,8 @@
     <thead>
         <tr>
             <th class="col-xs-1">#</th>
-            <th class="col-xs-3">{% trans "Person" %}</th>
-            <th class="col-xs-2"">{% trans "Roles" %}</th>
+            <th class="col-xs-4">{% trans "Person" %}</th>
+            <th class="col-xs-1">{% trans "Roles" %}</th>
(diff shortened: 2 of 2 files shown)


tristanlatr:
Hi @victorphoenix3,

Thanks for your contribution. 

It looks like you update the title only. 
I'suggest we have a button with text inside it, just so it's clear (right now it only appear on hover the element).

victorphoenix3:
@tristanlatr yes, but the button overflows onto the next line. Here is an attached screenshot if I use a button and without a button as well. What do you recommend?

Screenshot 2022-03-27 at 12 14 34 PM

tristanlatr:
Ok i see the kind of situation we’re having. Maybe it’s ok to have the text as a tooltip, but the default behavior only shows the tooltip after a couple seconds. We could add some JS so show the tooltip instantaneously, that would solve the issue I think. I’ve found some JS code that does that for another project that I’m working on here https://github.com/twisted/pydoctor/issues/524 so maybe we could do something similar? 

What do you think @victorphoenix3 ? Any insight @2ynn ?

Thanks

2ynn:
I don't really have a preference for the front-end to be honest. Also keep in mind that very few people should have the ` {% if perms.member.change_authuser %}` permission, so it would likely be trained core members.

@victorphoenix3 it might be relevant to adjust the column width ?
This is done in `view.html`
```
            <th class="col-xs-1">#</th>
            <th class="col-xs-3">{% trans "Person" %}</th>
            <th class="col-xs-2"">{% trans "Roles" %}</th>
            <th class="col-xs-2">{% trans "Harvests" %}</th>
            <th class="col-xs-2">{% trans "Properties" %}</th>
            <th class="col-xs-2">{% trans "Neighborhood" %}</th>
```
the numbers in `col-xs-*` classes must always amount to 12 in total. See bootstrap v3 documentation for more details

victorphoenix3:
I have added buttons and decreased the column width for `Roles`

Screenshot 2022-03-29 at 8 42 03 AM

tristanlatr:
> Also keep in mind that very few users should have the {% if perms.member.change_authuser %} permission, so it would likely be for trained core members only.

@2ynn you'r saying that the blue button should appear only on a few rows, not all of them ? If that's correct, the proposed look it very good. 

In any case, it looks OK for now, until we have more actions to do, maybe for a long-term solution we should think of adopting a unified row style for all list views and a ``...`` button that would reveal all actions. Like here we have already two actions for each user: register and edit... 🤔

2ynn:
@tristanlatr actually i personally don't think a `reset password` button should look that inviting and big. It's really not a button that should be used that often is it? I guess there's a confirmation pop up at least. For the register button though i'm okay if it looks big, it's good to see unregistered users and invite core members to register them.

No for the permission it depends on the logged in user so if the button is there it'll be on all rows.

tristanlatr:
> actually i personally don't think a reset password button should look that inviting and big.

True...

> For the register button though i'm okay if it looks big, it's good to see unregistered users and invite core members to register them.

Ok so let's do that for the short term and see how it goes. So no big button for the reset password, but only for the register button.

It's ok to adopt something for the moment and change it further later as we see fit. 

Talk to you later

victorphoenix3:
Updated

Screenshot 2022-03-30 at 7 04 50 PM

tristanlatr (approved):
Looks great! Thanks!

PR #143, opened 2022-03-28
[fix #32] Added success message if harvest status updated to "Succeeded"

### LINKED ISSUE #32: Kudos messages for pickeaders when harvest status is set to "Succeeded"
From: https://github.com/tiagovaz/saskatoon/issues/146

Ideas for text (this will need to bilingual):

“You’ve just led your "nth" fruit harvest this season! Thank you for supporting your community!”

May be in a green box somewhere, but not a pop-up.

mina-andajos:
fixes #32 
---------------------
After Editing Harvest on `/harvest/update/<pk>` and setting Harvest status to "Succeeded":

image
----------
On the success url, the success message changes to "You’ve just led your `harvest's pick-author successful number of harvests` fruit harvest this season! Thank you for supporting your community!":

image

### DIFF
--- a/saskatoon/harvest/views.py
+++ b/saskatoon/harvest/views.py
@@ -1,8 +1,10 @@
 from django.utils.translation import ugettext_lazy as _
+from django.utils import timezone
 from django.contrib import messages
 from django.contrib.messages.views import SuccessMessageMixin
 from django.contrib.auth.mixins import LoginRequiredMixin, PermissionRequiredMixin
 from django.contrib.auth.decorators import login_required
+from django.contrib.humanize.templatetags.humanize import ordinal
 
 from django.http import HttpResponseRedirect
 from django.urls import reverse_lazy
@@ -131,6 +133,21 @@ class HarvestUpdateView(LoginRequiredMixin, SuccessMessageMixin, UpdateView):
         context['title'] = _("Edit harvest")
         context['cancel_url'] = reverse_lazy('harvest-detail', kwargs={'pk': self.object.pk})
         return context
+
+    def get_success_message(self, cleaned_data) -> str:
+        harvests_as_pickleader = self.object.pick_leader.person.harvests_as_pickleader
+        harvests_this_year = harvests_as_pickleader.filter(
+            start_date__year=timezone.now().date().year
+        )
+        harvest_number: int = harvests_this_year.count()
+
+        success_message_harvest_successful: str = _(
+            "You’ve just led your {} fruit harvest! Thank you for supporting your community!"
+        ).format(ordinal(harvest_number))
+
+        if self.object.status == "Succeeded":
+            return success_message_harvest_successful % cleaned_data
+        return self.success_message % cleaned_data
 
     def get_success_url(self):
             return reverse_lazy('harvest-detail', kwargs={'pk': self.object.pk})


mina-andajos:
@2ynn changed the string formatting as you suggested. 👍

saskatoon/harvest/views.py
```diff
@@ -132,6 +133,18 @@ def get_context_data(self, **kwargs):
         context['cancel_url'] = reverse_lazy('harvest-detail', kwargs={'pk': self.object.pk})
         return context
 
+    def get_success_message(self, cleaned_data) -> str:
+        pick_leader_harvest_number: int = len(
+            self.object.pick_leader.person.harvests_as_pickleader
```
tristanlatr, line 138: Is this list contains harvests from past seasons ?

tristanlatr:
Thanks @mina-andajos,
I only have a little question regarding the season context, we should make sure that the picks led are in the current calendar year I think. Or maybe I'm missing (@2ynn)?

saskatoon/harvest/views.py
mina-andajos: I believe so yes, all the harvests the person has done. @tristanlatr 
```python
class Person:
    ...

    @property
    def harvests_as_pickleader(self):
        return Harvest.objects.filter(pick_leader=self.auth_user)
```

2ynn:
@tristanlatr @mina-andajos yes it's for all seasons, I'd simply remove "this season" from the kudo text. You could also use a timedelta of 1 year and filter queryset by harvest date

tristanlatr:
I think it's important to give the number per season. Some people have led like 40 picks...

mina-andajos:
> @tristanlatr @mina-andajos yes it's for all seasons, I'd simply remove "this season" from the kudo text. You could also use a timedelta of 1 year and filter queryset by harvest date
---------------
Which date in the `Harvest` model should I filter by? There are 4: `start_date, end_date, publication_date, creation_date `

```python
class Harvest(models.Model):
    ...

    start_date = models.DateTimeField(
        verbose_name=_("Start date"),
        blank=True,
        null=True
    )

    end_date = models.DateTimeField(
        verbose_name=_("End date"),
        blank=True,
        null=True
    )

    publication_date = models.DateTimeField(
        verbose_name=_("Publication date"),
        blank=True,
        null=True
    )

    ...

    creation_date = models.DateTimeField(
        verbose_name=_("Creation date"),
        auto_now=False,
        auto_now_add=True
    )
```

mina-andajos:
I have updated the commit. Used `start_date` for filtering, still haven't applied @tristanlatr's suggestion until I get a confirmation.
----------
Screenshot:
image

2ynn:
@mina-andajos `start_date` is good for filtering but I think my idea of a 1 year timedelta is not suited here. As @tristanlatr mentioned you want to look for harvests led during the current calendar year (btw if you look at the `start_date` filter in `harvest.filters.HarvestFilter` we basically look at the current year to define a "Season"). So something like `filter(start_date__year=timezone.now().date().year)` would be better.

mina-andajos:
@2ynn Thanks for your help. I felt something was off with the timedelta too.

tristanlatr:
It looks great, thanks a lot!

saskatoon/harvest/views.py
```diff
@@ -132,6 +134,21 @@ def get_context_data(self, **kwargs):
         context['cancel_url'] = reverse_lazy('harvest-detail', kwargs={'pk': self.object.pk})
         return context
 
+    def get_success_message(self, cleaned_data) -> str:
+        harvests_as_pickleader = self.object.pick_leader.person.harvests_as_pickleader
+        harvests_this_year = harvests_as_pickleader.filter(
+            start_date__year=timezone.now().date().year
+        )
```
2ynn, line 141: @mina-andajos you could use `.count()` after filtering, I think this would be more efficient than using `len()`

2ynn: here is relevant documentation https://docs.djangoproject.com/en/4.0/ref/models/querysets/#when-querysets-are-evaluated-1
see the Note in `len()` section

mina-andajos: Thanks for pointing this out

2ynn (approved):

Issue #123, opened 2022-03-18
Fix the season filter


tristanlatr:
Go to https://saskatoon-dev.lesfruitsdefendus.org/harvest/ and try to select another season that 2021, you will maybe see a pop with a datetime selector (like in harvest datetime scheduling).  

The plain dropdown list also shows on top of the datetime selector.

Should not be too hard to simply remove the datetime selector. 

![Capture d’écran, le 2022-03-18 à 16 43 42](https://user-images.githubusercontent.com/19967168/159080629-9d87c320-4cac-4f69-bc16-0c18e0540801.png)

2ynn:
see #146

PR #146, opened 2022-03-28
Fix for datetimepicker inside season filter on harvest page

farhanmasud:
Intended fix for [#123](https://github.com/LesFruitsDefendus/saskatoon-ng/issues/123)

The issue was the ID of the season filter field is "id_start_date" which is automatically generated by [django-crispy-forms](https://github.com/django-crispy-forms/django-crispy-forms) based on the field name and it's the same ID as the "Start date" field that's on the "Add a new harvest" form / page where the datetimepicker was originally intended for. 

To initiate the datetimepicker, the following code was included in `saskatoon/sitebase/templates/app/base/scripts.html` file - 

```
$("#id_start_date").datetimepicker({
     format: 'd/m/Y H:i',
 });
```

Since the above code is in a base template that other templates are inheriting from, it was also active on the "Harvests" page. Since it's not intended to be executed on this specific page, I have added a condition to check the URL of the current page using jQuery and if it's path is `/harvests/`, the datetimepicker is not executed anymore.

### DIFF
--- a/saskatoon/harvest/filters.py
+++ b/saskatoon/harvest/filters.py
@@ -19,7 +19,8 @@ class HarvestFilter(filters.FilterSet):
     seasons = list(set(seasons))
     seasons = sorted(seasons, key=lambda tup: tup[1])
 
-    start_date = filters.ChoiceFilter(
+    season = filters.ChoiceFilter(
+        field_name='start_date', 
         choices=seasons,
         label=_("Season"),
         lookup_expr='year',
@@ -55,13 +56,13 @@ class HarvestFilter(filters.FilterSet):
 
     class Meta:
         model = Harvest
-        fields = {
-        'status': ['exact'],
-        'pick_leader': ['exact'],
-        'trees': ['exact'],
-        'property__neighborhood': ['exact'],
-        'start_date': ['exact'],
-        }
+        fields = (
+            'status',
+            'pick_leader',
+            'trees',
+            'property__neighborhood',
+            'season',
+        )
 
 
 class PropertyFilter(filters.FilterSet):
--- a/saskatoon/harvest/api.py
+++ b/saskatoon/harvest/api.py
@@ -39,7 +39,7 @@ class HarvestViewset(LoginRequiredMixin, viewsets.ModelViewSet):
                         'property',
                         'about',
                         'status',
-                        'start_date')
+                        'season')
 
     # Harvest detail
     def retrieve(self, request, format='html', pk=None):


tristanlatr:
Thanks @farhanmasud,

> The issue was the ID of the season filter field is "id_start_date" which is automatically generated by [django-crispy-forms](https://github.com/django-crispy-forms/django-crispy-forms)

Do you know if it's possible to simply drop this id from being generated for this field?

Cause the proposed changes feels like a bit hacky, but maybe @2ynn feels differently? Are you sure it doesn't influence other inputs as well ? (I haven't tried it locally yet)

Thanks

farhanmasud:
@tristanlatr I'm checking this

farhanmasud:
Resetting last commit by `git reset --hard HEAD~1` and force pushing `git push --force` has closed my PR. I'm reopening it. Apologies for the inconvenience.

In this new commit, I have updated the code to change the ID from inside `HarvestViewset` class > `list` method by updating the filter form. The filter seems to be working as expected. Attaching a demo screen record below.

https://user-images.githubusercontent.com/43474661/160603634-c4be22ef-b195-4273-afa2-30adbd5681d3.mp4

farhanmasud:
@2ynn sure, I'll work on this custom method

saskatoon/harvest/filters.py
```diff
@@ -36,32 +38,41 @@ class HarvestFilter(filters.FilterSet):
...
         ),
         required=False,
+        lookup_expr="exact",
         help_text="",
     )
 
     trees = filters.ModelChoiceFilter(
         queryset=TreeType.objects.all(),
         label=_("Tree"),
+        lookup_expr="exact",
```
2ynn, line 48: `lookup_expr` defaults to "exact", it's not really needed to add them.

saskatoon/harvest/filters.py
```diff
@@ -36,32 +38,41 @@ class HarvestFilter(filters.FilterSet):
...
+            'trees',
+            'property__neighborhood',
+            'season',
+        )
+
+    def filter_season(self, queryset, name, value):
+        if value:
+            lookup = '__'.join([name, 'year'])
+            return queryset.filter(**{lookup: value})
+        return queryset
```
2ynn, line 75: is this working? name will be `season` but you want to look for `start_date__year`

saskatoon/harvest/filters.py
farhanmasud: These `exact` declarations were previously on the `fields` attribute under the `Meta` class. Since I had to remove them for the error mentioned in my comment, I have added them up with the each specific fields. Should I remove them?

saskatoon/harvest/filters.py
farhanmasud: Yes, I have tested it and it's working. The line `lookup = '__'.join([name, 'year'])` returns `start_date__year`

saskatoon/harvest/filters.py
```diff
@@ -19,15 +19,17 @@ class HarvestFilter(filters.FilterSet):
     seasons = list(set(seasons))
     seasons = sorted(seasons, key=lambda tup: tup[1])
 
-    start_date = filters.ChoiceFilter(
+    season = filters.ChoiceFilter(
+        field_name="start_date",
```
2ynn, line 23: @farhanmasud ok I didn't know you could add the `field_name`. Could we then get away without a custom method and simply go back to setting up `lookup_expr=year` ?

farhanmasud: Just checked and it's working without the custom method and using `lookup_expr=year` along with `field_name`. I'm updating it. Should I remove the `lookup_expr="exact",` from the other fields as well?

2ynn: Ok great, yes they're not needed you can remove them.

2ynn:
@farhanmasud please squash your commits when you're done. thanks!

tristanlatr:
Hello,

Tried it locally it works as expected. 

Thanks @farhanmasud, clean work.

I'm merging this PR.

farhanmasud:
Thanks a lot @2ynn and @tristanlatr for your guidance, really happy to be able to contribute

PR #151, opened 2022-03-30
Fix-some-translations

donaldte:

### DIFF
--- a/saskatoon/member/models.py
+++ b/saskatoon/member/models.py
@@ -10,7 +10,7 @@ from harvest.models import RequestForParticipation, Harvest, Property
 class AuthUserManager(BaseUserManager):
     def create_user(self, email, password=None):
         if not email:
-            raise ValueError('Users must have an email address')
+            raise ValueError(_('Users must have an email address'))
 
         user = self.model(email=self.normalize_email(email),
                           )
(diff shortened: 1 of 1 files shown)


tristanlatr (approved):
Looks good, thanks @donaldte!

2ynn:
@donaldte thanks for the changes! 

I believe the first message in `create_user` will only appear when calling `manage.py createsuperuser` from command line so a translation tag might not be relevant here but it doesn't  hurt to be consistent.

While you're at it would you mind changing the single quotes for double quotes? It's super nitty gritty but I personally would like to keep double quotes for strings and single quotes for dictionary or tuple keys. e.g.: `my_dict['key'] = "string"` or `fields = ['field1', 'field2']`, etc...
@tristanlatr are you okay with trying to stick to that rule, I don't think PEP8 recommends anything in particular.

Rerunning makemessages after adding additional `gettext_lazy()` tags should not delete existing translations otherwise this would be a nightmare (to be tested though). But yea we need to make sure to keep the .po files up to date so no time is spent translating text that is no longer relevant (e.g. moved or removed from the source code).

2ynn:
sure combining two kinds of quotes definitely makes sense in those cases.
Again I'm not talking about blocking contributions. My ideas behind uniformizing strings are about making it easier to find all translated strings at once by searching for `_("`  for example.

PR #152, opened 2022-03-30
fix some translation

donaldte:


saskatoon/harvest/admin.py
```diff
@@ -9,15 +9,16 @@
...
 class PropertyInline(admin.TabularInline):
     model = Property
     extra = 0
 
 class PersonInline(admin.TabularInline):
     model = RequestForParticipation
-    verbose_name = "Cueilleurs pour cette récolte"
-    verbose_name_plural = "Cueilleurs pour cette récolte"
+    verbose_name = _("Cueilleurs pour cette récolte")
+    verbose_name_plural = _("Cueilleurs pour cette récolte")
```
2ynn, line 21: It's actually spelled "Ceuilleur"
@tristanlatr Could we maybe evaluate how many strings are currently in French and marked for translations? I feel like it would make more sense to stick to one language in the source code and update translations.
@donaldte  I also feel like it would make sense to gather all your translation-related commits into a single pull request but no worries if that's too much of a hassle

tristanlatr: > Could we maybe evaluate how many strings are currently in French and marked for translations? I feel like it would make more sense to stick to one language in the source code and update translations accordingly.

not a priority imo, but we could hack something to read the .po files and detect the main language used for each lazy get text entry. But it sounds hard for little befits. Not sure what to think about this.

2ynn: I didn't find any other string marked for translation or  `verbose_name` that was in french. So if you don't mind I'd go with:
```
    verbose_name = _("Picker for this harvest")
    verbose_name_plural = _("Pickers for this harvest")
```

2ynn: and I'm sorry I have no idea why I thought the french version was misspelled, "Cueilleur" is right. But yea having it in english will at least leave all the decisions about inclusivity on the rosetta side and not in the source code.

donaldte: > ~It's actually spelled "Ceuilleur"~ @tristanlatr Could we maybe evaluate how many strings are currently in French and marked for translations? I feel like it would make more sense to stick to one language in the source code and update translations accordingly. @donaldte I also feel like it would make sense to gather all your translation-related commits into a single pull request but no worries if that's too much of a hassle

okey I gonna push all the translations in one commit.

2ynn (changes requested):
```
    verbose_name = _("Picker for this harvest")
    verbose_name_plural = _("Pickers for this harvest")
 ```

PR #153, opened 2022-03-30
fix some translation in admin's member file

donaldte:

PR #154, opened 2022-03-30
fix some tranlation in admin's member file

donaldte:


tristanlatr:
This PR is the same as #153 but targets the master branch, right ?

PR #162, opened 2022-04-01
Update INSTALL.md

khabdrick:
When a contributor clicks the copy button, the commands copy with the $ symbol so I just thought maybe it will be helpful if it is removed.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -25,9 +25,9 @@ Please follow each part of this documentation in order to run your own instance 
     For Ubuntu, Debian derivatives (for macOS, you should be able to use `pip`):
     ```
 
-    $ virtualenv venv
-    $ . venv/bin/activate
-    $(venv) pip3 install .
+    virtualenv venv
+    . venv/bin/activate
+    (venv) pip3 install .
     ```
 
     See `setup.py` for more details on the project's package requirements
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -25,9 +25,9 @@ Please follow each part of this documentation in order to run your own instance
     For Ubuntu, Debian derivatives (for macOS, you should be able to use `pip`):
     ```
 
-    $ virtualenv venv
-    $ . venv/bin/activate
-    $(venv) pip3 install .
+    virtualenv venv
+    . venv/bin/activate
+    (venv) pip3 install .
```
2ynn, line 30: ok I get your point but you wouldn't want to include `(venv)` in your pasted command. 
`$` and `$(venv)` are here to clarify when you need to be in the virtual environment. You mention a "copy button", are you talking about Github specifically? I don't think all contributors would use Github to read a markdown file. I don't personally.

2ynn: @khabdrick do you want to follow up on this or should I close this PR?

khabdrick: Ohhh sorry I missed this. If the change isn't necessary, you can go ahead.

khabdrick:
Ohh okay, no problem then. Was only suggesting.

PR #163, opened 2022-04-01
Update INSTALL.md

khabdrick:
The command to generate secrets wasn't working directly. I have added one that does.


tristanlatr (approved):
Hi @khabdrick, 

Thanks for your contribution.

It was working, but within a python interactive interpreter. 

But `python -c` works just as well. 

Thanks.

PR #164, opened 2022-04-01
added python-dotenv

khabdrick:
python-dotenv is used for testing but wasn't added to the `setup.py`. 

I added it in this pull request.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,6 +45,6 @@ setup(
         "django-rosetta==0.9.8"
     ],
     extras_require      =   {
-        'test': ['pytest', 'selenium', 'pytest-django', 'invoke', 'tox']
+        'test': ['pytest', 'selenium', 'pytest-django', 'invoke', 'tox', 'python-dotenv']
     },
 )


2ynn:
@khabdrick we are now using `django-dotenv` (see setup.py now updated in master) instead of `python-dotenv`.
Can you update your virtual environment requirements accordingly and see if changes are still required? Thanks!

khabdrick:
Hello @2ynn I compared the two documentations and found out that it's `python-dotenv` that is used for testing in this case, not `django-dotenv`.
https://github.com/jpadilla/django-dotenv#django-dotenv
https://pypi.org/project/python-dotenv/

Also found out that  `python-dotenv` won't work if `django-dotenv` is installed, so if this change is accepted I can update that.

khabdrick:
@2ynn @tristanlatr 
What do you think is the best way to move forward now?

2ynn:
@khabdrick is TESTDRIVER variable the only one you need from .env file in `conftest.py` ?
If so can you read it in `saskatoon.settings.py` using `django-dotenv` like other .env variables and then import it from there using `settings.TESTDRIVER` ?

khabdrick:
> @khabdrick is TESTDRIVER variable the only one you need from .env file in `conftest.py` ? If so can you read it in `saskatoon.settings.py` using `django-dotenv` like other .env variables and then import it from there using `settings.TESTDRIVER` ?

I think this will work.

tristanlatr:
> settings.TESTDRIVER

I don't think this variable exist. I'de suggest not create it because it's only for testing. 

Once you have imported the python module `saskatoon.settings` into `saskatoon.tests.conftest` the value of `TESTDRIVER` should be fetched with `os.getenv('TESTDRIVER')`.

tristanlatr:
> Same for DATABASES: can we also use the one from settings? I'm not sure why we need to override it in django_db_setup()

This is linked to the pytest-django setup that is currently not working...

khabdrick:
I don't know what to do now :(

tristanlatr:
> I don't know what to do now :(

Don’t worry about the broken tests, you can work on another issue that you find interesting. In the meantime, I’ll look at fixing the tests, but it’s not a great priority for now.

2ynn:
Ok I see what you mean now. But instead of importing the whole settings module could we just import the `dotenv` file then ? e.g.
```
from saskatoon.settings import dotenv
from dotenv import read_dotenv

TESTDRIVER = os.getenv('SASKATOON_TEST_WEBDRIVER')
```

2ynn:
Ok I see what you mean now. So if I understand well this should simply work:
```
import os
import pytest
import selenium.webdriver
from django.conf import settings

TESTDRIVER = os.getenv('SASKATOON_TEST_WEBDRIVER')
[...]
```

khabdrick:
> > I don't know what to do now :(
> 
> Don’t worry about the broken tests, you can work on another issue that you find interesting. In the meantime, I’ll look at fixing the tests, but it’s not a great priority for now.

Okay, no problem

PR #166, opened 2022-04-02
fix some translations

donaldte:
fix some translations in login.html file


2ynn:
I think your commit message could be more clear that it has to do with translations.

donaldte:
> I think your commit message could be more clear that it has to do with translations.

okay understood

tristanlatr (approved):
Looks good, thanks a lot @donaldte

PR #169, opened 2022-04-03
[WIP] Statistics page

mina-andajos:
I believe using serializers for `HarvestYield` would be best instead of using django views (since we want to user serializers everywhere according to https://github.com/LesFruitsDefendus/saskatoon-ng/discussions/144).


2ynn:
if you don't mind I'll let you rebase this one onto develop

saskatoon/harvest/api.py
```diff
@@ -223,3 +223,19 @@ def list(self, request, *args, **kwargs):
...
                          })
+
+class HarvestYieldViewset(LoginRequiredMixin, viewsets.ModelViewSet):
+    template_name = 'app/stats.html'
+    queryset = HarvestYield.objects.all().order_by('-id')
+    serializer_class = HarvestYieldSerializer
+    filter_backends = (filters.DjangoFilterBackend,)
+    filterset_class = HarvestYieldFilter
+    filterset_fields = '__all__'
+
```
2ynn, line 234: @mina-andajos ideally we should enforce custom permissions on this view, e.g. `permission_classes = [IsCoreOrAdmin]`
https://www.django-rest-framework.org/api-guide/permissions/#custom-permissions 
The `IsCoreOrAdmin` permission should subclass `IsAuthenticated` and make sure that the `request.user` belongs to either the `core` or `admin` group ( e.g. something like `request.user.groups.filter(name__in=["core", "admin"]).exists()` )
Let me know if you feel like implementing this, otherwise I can do it when you're done with you PR.
Thanks!

saskatoon/harvest/urls.py
```diff
@@ -79,9 +80,6 @@
          views.TemplateView.as_view(template_name='app/property_thanks.html'),
          name='property-thanks'),
 
-    path(r'stats/',
-         views.TemplateView.as_view(template_name='app/stats.html'),
-         name='statistics'),
```
2ynn, line 84: yes let's remove this!

saskatoon/harvest/api.py
mina-andajos: Thanks for pointing this out. I'll be implementing it.

mina-andajos:
sorry about the messy commit history. I will clean it up later.

saskatoon/harvest/api.py
mina-andajos: Had to create a `core` and an `admin` group since they didn't exist for me.

------------------------
### Screenshots:
1- not and in admin or core
image

2- core only
image

3- admin & core
image

2ynn: yea sorry the groups are loadded with `python saskatoon/manage.py saskatoon/fixtures/auth-group.json`

saskatoon/harvest/permissions.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework.permissions import IsAuthenticated
+from django.http import HttpRequest
+
+
+class IsCoreOrAdmin(IsAuthenticated):
+    message = "Viewing this page is restricted to core and admin users."
+
+    def has_permission(self, request: HttpRequest, view):
+        return request.user.groups.filter(name__in=["core", "admin"]).exists()
```
2ynn, line 9: @mina-andajos since we are subclassing `IsAuthenticated` it is important to first check the user is logged in, otherwise i believe you will get an error because Anonymous user has no groups.
e.g.:
```
        if super().has_permission(request, view):
            return request.user.groups.filter(name__in=["core", "admin"]).exists()
        return False
```

saskatoon/harvest/permissions.py
```diff
@@ -0,0 +1,9 @@
+from rest_framework.permissions import IsAuthenticated
+from django.http import HttpRequest
+
+
+class IsCoreOrAdmin(IsAuthenticated):
+    message = "Viewing this page is restricted to core and admin users."
```
2ynn, line 6: please let's mark this one for translation with `gettext_lazy()`

mina-andajos:
I replaced the HarvestYield Serializer and filter, took the approach of the old codebase (YAGNI).

so far it's the same functionality as the old stats view, I'm facing a problem with the routing so I'm working on it.
so far:
image

mina-andajos:
[StatsView old code for reference](https://github.com/tiagovaz/saskatoon/blob/e0a522b675594ba72c5aadbf918dd11c22939478/saskatoon/harvest/views.py#L736)
removed some redundant code, got rid of unused functions there, only used what's needed to form the context.

mina-andajos:
Edited the Viewset, added a placeholder template to show results.

screenshots:
showing all seasons:
image

showing a specific season:
image

when entering an invalid season:
image
----------
@tristanlatr @2ynn could you please review the backend code? if all's good I'll start on making the template right away 👍

mina-andajos:
Progress so far in the UI:

image

saskatoon/harvest/api.py
```diff
@@ -221,3 +224,188 @@ def list(self, request, *args, **kwargs):
...
+        )
+
+    def get_highlights(self):
+        total_weight = int(
+            HarvestYield.objects.aggregate(Sum("total_in_lb")).get("total_in_lb__sum")
+        )
+        total_harvests = Harvest.objects.count()
+        total_pickers = HarvestYield.objects.values("recipient").distinct().count()
+        total_beneficiaries = Organization.objects.filter(is_beneficiary=True).count()
+
```
2ynn, line 293: @mina-andajos I'm a bit confused by the effect of the `season` filter. When we pass a season as kwarg, we should expect stats to be filtered over that season right?

saskatoon/harvest/api.py
```diff
@@ -221,3 +224,188 @@ def list(self, request, *args, **kwargs):
...
+    # TODO Use the original seasons ChoiceField in dropdown
+    seasons = list(map(lambda x: x[0], filterset_class.seasons))
+
+    def list(self, request, format="html", *args, **kwargs):
+        """returns statistics on harvests for all seasons or a specific season"""
+
+        season_parameter = self.kwargs.get("season")
+        if season_parameter is not None:
+            if season_parameter in self.seasons:
+                self.queryset.filter(start_date__year=season_parameter)
```
2ynn, line 257: what is this line doing? I don't think you are actually modyfing the queryset here?
Have you looked at the `get_queryset` method? 
But either way I am not sure I understand where you are re-using `self.queryset` to compute your stats

saskatoon/harvest/api.py
```diff
@@ -221,3 +224,188 @@ def list(self, request, *args, **kwargs):
...
+
+        response = super(StatsViewset, self).list(request, *args, **kwargs)
+        if request.accepted_renderer.format == "json":
+            return Response(response.data)
+
+        highlights = self.get_highlights()
+        total_fruit = self.get_total_weight_per_fruit()
+        total_neighborhood = self.get_total_weight_per_neighborhood()
+        total_beneficiary = self.get_total_weight_per_beneficiary()
+        total_picker = self.get_total_weight_per_picker()
```
2ynn, line 269: I feel like all these functions should take the `season` as parameter and filter their querysets accordingly

saskatoon/harvest/api.py
```diff
@@ -221,3 +224,188 @@ def list(self, request, *args, **kwargs):
                                  'title': _("New Person")
                                  }
                          })
+
+
+class StatsViewset(LoginRequiredMixin, viewsets.ModelViewSet):
```
2ynn, line 229: Also the choice of using a ModelViewSet here might be questionable since you are not actually displaying a list of (filtered) harvests and you seem to only need a single view (with a kwarg argument). ViewSets are usually used to implement all CRUD operations. I feel like a GenericAPIView would be enough in your case.

mina-andajos:
@2ynn Thanks for pointing out the fixes needed for the codebase.
--------------
### I have applied the following:
1. inherited `APIView` since I didn't need to get a queryset.
2. added season filtering the `self.get.total_weight()` functions
3. refactored code related to the season parameter's kwarg
4. fixed a bug in the template's drop menu
-------------
please check the PR message for new screenshots.

saskatoon/harvest/api.py
```diff
@@ -221,3 +224,281 @@ def list(self, request, *args, **kwargs):
...
+                    .filter(harvest__property__neighborhood=neighborhood)
+                    .filter(harvest__start_date__year=season)
+                    .aggregate(Sum("total_in_lb"))
+                )
+                total_harvests = (
+                    Harvest.objects.filter(status="Succeeded")
+                    .filter(property__neighborhood=neighborhood)
+                    .filter(start_date__year=season)
+                    .count()
+                )
```
2ynn, line 356: @mina-andajos with hindsight I don't think passing the season as parameter is the most efficient, looks like we are filtering the same objects multiple times. Wouldn't it better to fetch the `Harvest` and `HarvestYield` querysets based on season at the top and pass these into the separate functions for further filtering? This way you shouldn't even have to check if season='all' everytime ?

mina-andajos: thanks for mentioning this i was looking for a way to refactor the functions used in the old saskatoon codebase. that's why I tried to add `self.queryset` before. I'll implement this right away.

saskatoon/harvest/api.py
```diff
@@ -122,7 +125,6 @@ def retrieve(self, request, format='html', pk=None):
             return response
         # default request format is html:
         similar = get_similar_properties(self.get_object())
-        print("similar", similar)
```
2ynn, line 125: please rebase onto develop, I think I removed this already

saskatoon/harvest/api.py
```diff
@@ -221,3 +223,173 @@ def list(self, request, *args, **kwargs):
...
+
+    template_name = "app/stats.html"
+    permission_classes = [IsCoreOrAdmin]
+    # ? seasons that will be shown in the dropdown list
+    seasons = list(map(lambda x: x[0], HarvestFilter.seasons)) + ["all"]
+
+    def get(self, request, format="html", *args, **kwargs):
+        """returns statistics on harvests for all seasons or a specific season"""
+
+        season = self.kwargs.get("season", "all")
```
2ynn, line 238: ok I think using a slug parameter passed as kwargs for `season` works. In terms of scalability (e.g. if one day we decide we want to also filter by city or other fields) I still think using a `GenericAPIView` (which can handle queryset / filters) could be suited. For example you could reuse the season filter defined in HarvestFilter and retrieve the filtered queryset with something like this:

```
self.harvest_queryset = self.filter_queryset(self.get_queryset()) 
self.harvest_yield_queryset = HarvestYield.filter(harvest__in=harvest_qs)

```
and work from these two querysets in your functions.
That being said it might not be as straightforward because you are not working with a single model (Harvest + HarvestYield) so my comment might not be accurate. I'll let you evaluate if you think it's worth doing it, otherwise absolutely no worries if it works as such!

tristanlatr:
Hi @mina-andajos could you look into the following error? The CI is supposed to be green now :) 

2022-04-16T18:19:15.3181159Z ##[group]Run saskatoon/fixtures/init
2022-04-16T18:19:15.3181667Z [36;1msaskatoon/fixtures/init[0m
2022-04-16T18:19:15.3237129Z shell: /usr/bin/bash -e {0}
2022-04-16T18:19:15.3237358Z env:
2022-04-16T18:19:15.3237630Z   pythonLocation: /opt/hostedtoolcache/Python/3.8.12/x64
2022-04-16T18:19:15.3237959Z   LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.8.12/x64/lib
2022-04-16T18:19:15.3238233Z ##[endgroup]
2022-04-16T18:19:17.1742421Z Operations to perform:
2022-04-16T18:19:17.1743214Z   Apply all migrations: admin, auth, contenttypes, harvest, member, sessions
2022-04-16T18:19:17.1743929Z Running migrations:
2022-04-16T18:19:17.1744241Z   No migrations to apply.
2022-04-16T18:19:17.2614130Z Loading auth-group
2022-04-16T18:19:19.0639458Z Traceback (most recent call last):
2022-04-16T18:19:19.0640343Z   File "../manage.py", line 21, in <module>
2022-04-16T18:19:19.0641026Z     main()
2022-04-16T18:19:19.0641398Z   File "../manage.py", line 17, in main
2022-04-16T18:19:19.0641788Z     execute_from_command_line(sys.argv)
2022-04-16T18:19:19.0642793Z   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
2022-04-16T18:19:19.0643206Z     utility.execute()
2022-04-16T18:19:19.0643861Z   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/django/core/management/__init__.py", line 413, in execute
2022-04-16T18:19:19.0644340Z     self.fetch_command(subcommand).run_from_argv(self.argv)
2022-04-16T18:19:19.0645034Z   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/django/core/management/base.py", line 354, in run_from_argv
2022-04-16T18:19:19.0645477Z     self.execute(*args, **cmd_options)
2022-04-16T18:19:19.0646121Z   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/django/core/management/base.py", line 393, in execute
2022-04-16T18:19:19.0646489Z     self.check()
2022-04-16T18:19:19.0647096Z   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/django/core/management/base.py", line 419, in check
2022-04-16T18:19:19.0647497Z     all_issues = checks.run_checks(
2022-04-16T18:19:19.0648177Z   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/django/core/checks/registry.py", line 76, in run_checks
2022-04-16T18:19:19.0648831Z     new_errors = check(app_configs=app_configs, databases=databases)
2022-04-16T18:19:19.0649508Z   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/django/core/checks/urls.py", line 13, in check_url_config
2022-04-16T18:19:19.0649912Z     return check_resolver(resolver)
2022-04-16T18:19:19.0650804Z   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/django/core/checks/urls.py", line 23, in check_resolver
2022-04-16T18:19:19.0651155Z     return check_method()
2022-04-16T18:19:19.0651753Z   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/django/urls/resolvers.py", line 416, in check
2022-04-16T18:19:19.0652155Z     for pattern in self.url_patterns:
2022-04-16T18:19:19.0652772Z   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/django/utils/functional.py", line 48, in __get__
2022-04-16T18:19:19.0653206Z     res = instance.__dict__[self.name] = self.func(instance)
2022-04-16T18:19:19.0653853Z   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/django/urls/resolvers.py", line 602, in url_patterns
2022-04-16T18:19:19.0654349Z     patterns = getattr(self.urlconf_module, "urlpatterns", self.urlconf_module)
2022-04-16T18:19:19.0655022Z   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/django/utils/functional.py", line 48, in __get__
2022-04-16T18:19:19.0655468Z     res = instance.__dict__[self.name] = self.func(instance)
2022-04-16T18:19:19.0656293Z   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/django/urls/resolvers.py", line 595, in urlconf_module
2022-04-16T18:19:19.0657246Z     return import_module(self.urlconf_name)
2022-04-16T18:19:19.0658103Z   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/importlib/__init__.py", line 127, in import_module
2022-04-16T18:19:19.0658707Z     return _bootstrap._gcd_import(name[level:], package, level)
2022-04-16T18:19:19.0659115Z   File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
2022-04-16T18:19:19.0659518Z   File "<frozen importlib._bootstrap>", line 991, in _find_and_load
2022-04-16T18:19:19.0659949Z   File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
2022-04-16T18:19:19.0660372Z   File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
2022-04-16T18:19:19.0660796Z   File "<frozen importlib._bootstrap_external>", line 843, in exec_module
2022-04-16T18:19:19.0661841Z   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
2022-04-16T18:19:19.0662745Z   File "/home/runner/work/saskatoon-ng/saskatoon-ng/saskatoon/saskatoon/urls.py", line 9, in <module>
2022-04-16T18:19:19.0663530Z     path('', include('harvest.urls')),
2022-04-16T18:19:19.0664392Z   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/django/urls/conf.py", line 34, in include
2022-04-16T18:19:19.0664824Z     urlconf_module = import_module(urlconf_module)
2022-04-16T18:19:19.0665317Z   File "/opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/importlib/__init__.py", line 127, in import_module
2022-04-16T18:19:19.0665744Z     return _bootstrap._gcd_import(name[level:], package, level)
2022-04-16T18:19:19.0666165Z   File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
2022-04-16T18:19:19.0666583Z   File "<frozen importlib._bootstrap>", line 991, in _find_and_load
2022-04-16T18:19:19.0667032Z   File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
2022-04-16T18:19:19.0667459Z   File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
2022-04-16T18:19:19.0667892Z   File "<frozen importlib._bootstrap_external>", line 843, in exec_module
2022-04-16T18:19:19.0668342Z   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
2022-04-16T18:19:19.0668968Z   File "/home/runner/work/saskatoon-ng/saskatoon-ng/saskatoon/harvest/urls.py", line 5, in <module>
2022-04-16T18:19:19.0669387Z     from harvest import api, views, autocomplete
2022-04-16T18:19:19.0669978Z   File "/home/runner/work/saskatoon-ng/saskatoon-ng/saskatoon/harvest/api.py", line 229, in <module>
2022-04-16T18:19:19.0670413Z     class StatsView(LoginRequiredMixin, views.APIView):
2022-04-16T18:19:19.0671023Z   File "/home/runner/work/saskatoon-ng/saskatoon-ng/saskatoon/harvest/api.py", line 234, in StatsView
2022-04-16T18:19:19.0671486Z     seasons = list(map(lambda x: x[0], HarvestFilter.seasons)) + ["all"]
2022-04-16T18:19:19.0672213Z AttributeError: type object 'HarvestFilter' has no attribute 'seasons'

mina-andajos:
hi @tristanlatr
I have rebased develop onto the branch. I'm just refactoring the code based on @2ynn's suggestion and I will force push right away. thanks.

mina-andajos:
@2ynn hi
Is this how you wanted to refactor the View? What took me this long is that I was trying to abstract the view a bit more by inheriting `ListAPIView` instead of `GenericAPIView`. But I didn't have much luck doing that. Do you have another suggestions or advise regarding that?

saskatoon/harvest/api.py
```diff
@@ -220,3 +224,185 @@ def list(self, request, *args, **kwargs):
...
+        if season not in self.seasons:
+            raise Http404("Season Not found")
+
+        if season == "all":
+            self.harvest_queryset = self.filter_queryset(self.get_queryset())
+            self.harvest_yield_queryset = HarvestYield.objects.filter(harvest__in=self.harvest_queryset)
+
+        else:
+            self.harvest_queryset = self.filter_queryset(self.get_queryset()).filter(
```
2ynn, line 251: What I had in mind would remove the need for passing a URL parameter manually (i.e. you would use the `season` filter from HarvestFilter REST would then handle the url parameter `?season=<choice>` for us.). Using `ListAPIView` instead of `GenericAPIView` might actually be a good idea!
Basically `self.filter_queryset(self.get_queryset())` would already contain the information about the season (all successful harvests if the filter is not applied). 
I think what would help you to troubleshoot would be to the the REST filter form (which you probably don't access to because you have already defined the html `template_name`.

saskatoon/harvest/urls.py
```diff
@@ -84,9 +84,9 @@
...
+    re_path(r'^stats/(?P<season>\w+)/$', api.StatsView.as_view(),name='statistics'),
```
2ynn, line 87: here you wouldn't need to define a slug manually, REST filter can handle that for you

2ynn (commented):
it's getting there!

mina-andajos:
@2ynn 
I've looked into [this tutorial](https://tech.serhatteker.com/post/2022-04/django-rest-filtering-tutorial-part-4/), and I could add code similar to this to `get_queryset()` for when we need to extend the filtering:
```py
    def get_queryset(self):
        qp = self.request.query_params

        param_list = ("first_name", "last_name", "name", "username")
        is_desired_param = any([param in qp for param in param_list])
        if len(qp) > 0 and not is_desired_param:
            return Author.objects.none()

        if (name := qp.get("name", None)) is not None:
            return Author.objects.filter(
                Q(first_name__icontains=name)
                | Q(last_name__icontains=name)
                | Q(user__name__icontains=name)
                | Q(user__username__icontains=name)
            )

        return super().get_queryset()
```

mina-andajos:
> @mina-andajos, if you le using a syntax like this:`if (name := qp.get("name", None)) is not None:`, please add `python_requires='>=3.8', ` param to the `setup.py` to be sure no one installs Saskatoon on python 3.7 (this would make python crash with  syntax error while evaluating this code)
> 
> Thanks :)

thanks for the headsup. I don't think it'll be worth it to use the walrus operator then if it will make problems for other devs.

tristanlatr:
> I don't think it'll be worth it to use the walrus operator then if it will make problems for other devs.

True, but also I think when I think about it, I feel like the developers secteur should be closest as possible as the prod setup. So I would be even more radical and use `python_requires='==3.8'`, and update this value when we’re updating the python version on the production server. With the appropriate testing before if course.

But this can be done in another PR.

saskatoon/harvest/api.py
```diff
@@ -220,3 +224,176 @@ def list(self, request, *args, **kwargs):
...
+    queryset = Harvest.objects.filter(status="Succeeded")
+    filter_backends = (filters.DjangoFilterBackend,)
+    filterset_class = HarvestFilter
+    filterset_fields = ('status', 'season')
+    permission_classes = [IsCoreOrAdmin]
+
+    def get_queryset(self):
+        self.season = self.kwargs.get('season')
+        if self.season is not None:
+            return self.queryset.filter(start_date__year=self.season)
```
2ynn, line 240: again, the `season` field in HarvestFilter should already do this for you when calling `stats/?season=<year>`

saskatoon/harvest/urls.py
```diff
@@ -84,9 +84,9 @@
          views.TemplateView.as_view(template_name='app/property_thanks.html'),
          name='property-thanks'),
 
-    path(r'stats/',
-         views.TemplateView.as_view(template_name='app/stats.html'),
-         name='statistics'),
+    path('stats/', api.StatsView.as_view(), name='statistics'),
+
+    path('stats/<int:season>/', api.StatsView.as_view(), name='statistics'),
```
2ynn, line 89: I believe `path('stats/', api.StatsView.as_view(), name='statistics')` would be enough and the season year could be passed using the `season` field from HarvestFilter with `stats/?season=2022`

2ynn (commented):
Sorry if I had not been very clear about removing the slug from the URL and using the HarvestFilter instead. It's really not a big deal. I can't test things locally right now though so we could either decide to go ahead with this PR or I could create a specific branch for this and see if I can help with refactoring the view when i get a chance next week.
Thanks for all this!

saskatoon/harvest/urls.py
mina-andajos: @2ynn hi, I'm now using url parameters as you advised. 

image

saskatoon/harvest/api.py
```diff
@@ -122,6 +125,7 @@ def retrieve(self, request, format='html', pk=None):
             return response
 
         # default request format is html:
+        similar = get_similar_properties(self.get_object())
```
2ynn, line 128: this line probably slipped back in unintentionally

saskatoon/harvest/api.py
```diff
@@ -220,3 +224,176 @@ def list(self, request, *args, **kwargs):
...
+    filter_backends = (filters.DjangoFilterBackend,)
+    filterset_class = HarvestFilter
+    filterset_fields = ('status', 'season')
+    permission_classes = [IsCoreOrAdmin]
+
+    def get_queryset(self):
+        self.season = self.request.query_params.get('season')
+        if self.season is not None:
+            return self.queryset.filter(start_date__year=self.season)
+        return super().get_queryset()
```
2ynn, line 241: Do you still need this? The `season` filter called by passing `?season=2021` to the url should already take action when calling `self.filter_queryset()` no ? This was my point about not filtering the queryset manually but simply letting Rest handling it since we already have a `season` field in `HarvestFilter`. Let me know if I'm missing something.

mina-andajos: I tried that yeah. it gave me this error: `'StatsView' object has no attribute 'season'`.
I followed how the [documentation](https://www.django-rest-framework.org/api-guide/filtering/#filtering-against-query-parameters) used filtering against query parameters.

saskatoon/harvest/api.py
```diff
@@ -220,3 +223,176 @@ def list(self, request, *args, **kwargs):
...
+        return len(total_number_beneficiaries)
+
+    @property
+    def seasons(self):
+        """seasons that will be shown in the dropdown list"""
+        seasons = []
+        for harvest in self.queryset.all():
+            if harvest.start_date is not None:
+                seasons.append(harvest.start_date.year)
+        return sorted(list(set(seasons)))
```
2ynn, line 398: I think going through all harvests is overkill for what we need, I have already changed the way we get the `season.choices` in `HarvestFilter` to avoid this,

Please try using `HarvestFilter.YEARS` instead:
```
        return Response(
            {
                "season": self.request.query_params.get('season'),
                "seasons": HarvestFilter.YEARS,
                ...
```

While you're at it I would appreciate if you could modify the season filter range in `harvest.filters` so we don't have to hardcode the current year:

```
from datetime import datetime

SEASON_FILTER_RANGE = (2016, datetime.now().year)

```
Thanks!

2ynn: That being said you will now run into issues for years where no there were no harvests:
```
Exception Type: | TypeError
-- | --
int() argument must be a string, a bytes-like object or a number, not 'NoneType'
saskatoon/harvest/api.py, line 256, in get_highlights
```
But I think we should cover the case were `self.get_queryset()` returns an empty queryset (e.g. if we decide to add more filters further down the road it could be possible to get no results)

mina-andajos: > That being said you will now run into issues for years where no there were no harvests:
> 
> ```
> Exception Type: | TypeError
> -- | --
> int() argument must be a string, a bytes-like object or a number, not 'NoneType'
> saskatoon/harvest/api.py, line 256, in get_highlights
> ```
> 
> But I think we should cover the case were `self.get_queryset()` returns an empty queryset (e.g. if we decide to add more filters further down the road it could be possible to get no results)

@2ynn
I'm thinking of adding a red error message and showing all stats. What do you think?

2ynn: @mina-andajos sure that works, showing no stats at all would also make sense.

BTW I've already pushed the change for YEARS here, should merge soon:
https://github.com/LesFruitsDefendus/saskatoon-ng/pull/286/commits/b937bd1f290f34471c14cb7395e36ef2c6e25e42

mina-andajos:
@2ynn hi,
I have made it that if after filtering using query parameters and the queryset is empty, display a queryset of all harvests without filtering and an error message. I don't think this implementation is clean however so any suggestions would be appreciated.

image

saskatoon/harvest/api.py
```diff
@@ -220,3 +223,170 @@ def list(self, request, *args, **kwargs):
...
+
+        self.harvest_queryset = self.filter_queryset(self.get_queryset())
+        #? Show error message and display all stats if the queryset is empty
+        if not self.harvest_queryset:
+            self.harvest_queryset = self.get_queryset()
+            messages.error(self.request, error_message, "danger")
+            #? display 'All' on the dropdown button instead of the specified season to avoid confusion
+            season = 'All'
+
+        self.harvest_yield_queryset = HarvestYield.objects.filter(harvest__in=self.harvest_queryset)
```
2ynn, line 249: First I think it'd maybe make more sense to define error_message under the if statement:
 ```
        self.harvest_queryset = self.filter_queryset(self.get_queryset())
        if not self.harvest_queryset:
            self.harvest_queryset = self.get_queryset()
            error_message = f"No harvests were found for {season=}. Displaying all statistics..."
            messages.error(self.request, error_message, "danger")
            season = 'All'
 ```

But then again your queryset could also be empty because of the `status` field available in your filter, then your message wouldn't really be accurate (same for other filters we might want to add in the future).
Also what if for some reason `self.get_queryset()` is empty (e.g. fresh database) ? It sure is an edge case but getting a server error could be avoided.
      
Are the `aggregate(Sum("field_name"))` bits throwing errors on empty querysets? 
      
I personally think it'd make more sense to present an empty page, that is that each of your function should handle  an empty queryset and return 0.  
Or you could also simply return 0 for all of them right away, definitely not the most elegant though, e.g.:

```
      if not self.harvest_queryset:
      
          no_highlights = {"total_harvests": 0,
                               "total_pickers": 0,
                                "total_beneficiaries": 0,
                                "total_weight": 0 }
                                
              return Response(
            {
                "season": season,
                "seasons": self.filterset_class.YEARS,
                "highlights": no_hightlights,
                "total_fruit": 0,
                "total_neighborhood": 0,
                "total_beneficiary": 0,
                "total_picker": 0,
            }
        )
```
     
If it's too much of a hassle we can keep your solution for now

mina-andajos: Thanks for pointing all this out, very valid points!
>Are the aggregate(Sum("field_name")) bits throwing errors on empty querysets?

Yes that's where I get an error:
```sh
Exception Type: TypeError at /stats/
Exception Value: int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```
--------
For empty data that's what gets displayed in the tables:

image

Is that good, or would adding a single row with `N/A` data would be more good looking?.

2ynn: yes that's totally fine! Why do you still get non null values in all other tables though?

mina-andajos: I've added this harvest through the admin panel. It probably let me add the harvest without adding what trees the picker have picked.

mina-andajos: image
It happens because I didn't add a person as a recipient.

image

mina-andajos:
@2ynn hi
I have edited the functions to return empty data if `harvest_queryset` is empty.
Output now:
image

Issue #167, opened 2022-04-03
Clarify branches and pull requests bases


2ynn:
@tristanlatr if we are using master branch to target pull requests, what is now the use for the develop branch?

Why not keep on merging pull requests into develop first? When they are tested on dev server we merge them to master.
Then master gets pushed to prod server. And as usual I would still push to prod branch from dev server so we always know what is currently up.
It's confusing now because all we have for new-generation is a "dev" server but I think it will become our production server soon and we have to eventually setup a second one.

Other option is we don't really use a develop branch anymore and just use master, but we still need some sort of a workflow to test things before they are pushed live. I feel keeping the develop branch would be safer, i don't always have time to test branches locally and sometimes i just look at the diffs.

I think it'd be good clarifying this in a CONTRIBUTING.md file

mina-andajos:
Yeah I was wondering the same 💯. Maybe adhering to the [git flow](https://youtu.be/aNJTAYwYIE8) feels like the safest option.

tristanlatr:
Hi,

> It's confusing now because all we have for new-generation is a "dev" server but I think it will become our production server soon and we have to eventually setup a second one.

We should figure out the required DNS changes.

> I feel keeping the develop branch would be safer,

Whatever works, but if we're not targeting pull requests to the main (which is master in our case), is should be indeed clear in the CONTRIBUTING file. Because Github UI targets the PR by default to the main branch. I think sticking to this is best. I'd suggest we update `develop` even to `master` and then make it the main branch. Then PRs created will automatically target the `develop` branch. Then we use `master`, `prod`, `testing`, `qa`  or whatever else for our QA -> Production flow. 

Tell me what you think,

Tristan

2ynn:
Ok I just made `develop` equal to `master` and set `develop` as default branch in GitHub.
I've also added a rule to prevent direct pushes to the master branch.
I went in most approved pull requests and changed the base branch to develop. I'll merge them shortly.

Thanks for the link @mina-andajos . I think for small fixes we can merge into `develop` and for bigger features we can start dedicated branches (e.g. I will start one for the PublicPropertyForm) so contributors can merge into them before we integrate into develop.

I will start a draft for a CONTRIBUTING.md file.
thanks

PR #170, opened 2022-04-03
[#167] add contribution guidelines to CONTRIBUTING.md

2ynn:
@tristanlatr @mina-andajos feel free to edit CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,43 @@
+# Contributing to saskatoon-ng
+
+Contributions are more than welcome!
+
+## Setup
+
+Please follow instructions in `INSTALL.md` to set up your local development environment.
+
+
+## Branches
+
+Unless instructed otherwise by maintainers all pull requests should target the `develop` branch.
+
+If your branch directly addresses an opened `Issue`, you may name your branch using the following scheme: `Fix-<Issue#>-A_short_description`
+
+Maintainers may create specific `Feature-<Issue#>-A_short_description` branches for larger issues so contributors can merge their patches into these branches before they get integrated into `develop`.
+
+
+## Pull requests
+
+Please reference the Issue number(s) your contribution is addressing, either in the Pull Request's title or in the first comment.
+
+To ease the review process, you may include screenshots, links to routes you've altered and a description of what was modified and why (e.g. in a bullet points form).
+
+If your pull request is still a *Work in Progress*, please mark it as `Draft`.
+
+
+## Commits
+
+If possible, squash your commits into a minimal number of relevant commits before finalizing your pull request. If you are unsure how to do it but are interested, feel free to ask for help.
+
+Although we don't have strict guidelines regarding commit messages, we recommend taking a look at [this practical guide](https://www.freecodecamp.org/news/writing-good-commit-messages-a-practical-guide/) to write better commits.
+
+If a commit message directly addresses an `Issue` it might be relevant to include the issue number in the commit message, e.g.: `[#119] Add postal_code field to PropertyAdmin list_display`.
+
+
+## Migrations
+
+If you alter any `<app>.models.py` file, make sure make sure to run `python manage.py makemigrations <app>** to include migrations along with the changes. 
+
+**Please only submit one migration file per app per pull request.** If you are not sure how to achieve this ask us for help and we will guide you through.
+
+Thanks!


2ynn:
I think a naming convention for Pull Request titles could be good to add as well

CONTRIBUTING.md
```diff
@@ -0,0 +1,32 @@
...
+If your branch directly addresses an opened `Issue`, please name your branch using the following scheme: `Fix-<Issue#>-A_short_description`
+
+Maintainers may create specific `Feature-<Issue#>-A_short_description` branches for larger issues so contributors can merge their patches into these branches before they get integrated into `develop`.
+
+
+## Commits
+
+Before finalizing your pull request, squash your commits into a minimal number of relevant commits.
+
+> TODO: include some commit messages guidelines (or link to a reference)
```
mina-andajos, line 23: @2ynn 
How about using imperative form in commit messages? Like `Fix some-bug` instead of `Fixed`. As that's how git makes commit messages when using git merge, same for github iirc.

Maybe if it's on this form: `Add/Refactor/Test/Docs/Remove commit-message`.
You and tristan write commits in different ways though so it depends on your convention.

Maybe also **not** use `[#issue-number] commit-message`, because we can reference issues in PRs.

You can check [this freecodecamp article](https://www.freecodecamp.org/news/writing-good-commit-messages-a-practical-guide/) too if would like to. There are some useful guidelines there too.

Tell me what you think of my suggestion and I'll add some edits.

mina-andajos:
Also a suggestion, just a preference of mine. How about we add a pre-commit hook to format the python code? Just to make everything look coherent. I can add something like [this](https://github.com/psf/black/blob/main/.pre-commit-hooks.yaml) and reference a guide (like [this one](https://www.youtube.com/watch?v=psjz6rwzMdk) on how to set it up for contributors.

CONTRIBUTING.md
2ynn: oh yes my commit messages are terrible, i definitely need guidelines ;) yes imperative form is good! 
ok for not including issue number in the PR title and referencing it in the comment section. To me having the issue number in commits though could help when looking back at the git log history. But maybe it doesn't make sense.

tristanlatr:
In my opinion, pull requests names, branch names and commit messages are largely aesthetic, and should not be constrained at this point of development.

What is important though, is having reviews and approvals all all PR that are open before merging them. Linking the issue that the PR is closing is also useful.

CONTRIBUTING.md
mina-andajos: Also how about adding [WIP] before a PR's name to let you know it's not finished yet and tell the contributors to mention you, share screenshots (I could edit the PR template to include maybe screenshots, links to pages, bullet points to what they did exactly, etc...)

2ynn:
I don't think is just aesthetics, at least on this repo (agree about not needing to put constraints on contributers' branch names on their own forks).  But for example when we need to clean up old branches it would be useful to quickly see which issue they addressed.

@mina-andajos I'd be down to switch to black but I've never set it up, i still use `yapf` as a general guideline and I don't automatically format on save.

2ynn:
> What is important though, is having reviews and approvals on all PR that are open before merging them. Linking the issue that the PR is closing is also useful.

@tristanlatr ok I added a protection on develop branch so all pull request need at least one approval

CONTRIBUTING.md
2ynn: can we simply use the `Draft` functionality for marking PRs as not ready?

mina-andajos:
> @mina-andajos I'd be down to switch to black but I've never set it up, i still use `yapf` as a general guideline and I don't automatically format on save.

We could setup [yapf](https://github.com/google/yapf/blob/main/plugins/pre-commit.sh) instead if that's what you use here. It'll just format all the staged changes automatically before committing. Whether the contributor use it in their editor or not. I can add the pre-commit and write instructions for contributors to set it up.

2ynn:
> We could setup [yapf](https://github.com/google/yapf/blob/main/plugins/pre-commit.sh) instead if that's what you use here. It'll just format all the staged changes automatically before committing. Whether the contributor use it in their editor or not. I can add the pre-commit and write instructions for contributors to set it up.

We haven't agreed on any formatters so far, so I'd be okay testing black. And yes as long as it only formats the actual changes and not the whole files I think that's good (not that we don't need general reformatting but let's not mess with the whole codebase for now)

CONTRIBUTING.md
tristanlatr: Yes using draft sounds better.

tristanlatr (approved):
Great guidelines, thanks.

mina-andajos (approved):
Thanks for your efforts.

PR #177, opened 2022-04-03
some harvest translations

tristanlatr:
Just 20mn of translation work on harvest app. 

For now, I did not fill one translation: "Canceled by picker". 

I think that we should adopt a strategy and use épicène from when possible to refer to participants. 

So "Canceled by picker" could become "Annulé par le/la cueilleuse.eur" in french, but, depending on the context we might just use "Annulé". Since écriture inclusive is not very standardized yet, we should adopt some clear guidelines before going forward.


2ynn:
yea sure would like to be as inclusive as possible. Maybe we can find workarounds when it becomes too "convoluted" , e.g. "Bénévole a annulé" ?

Superpat:
Hi @tristanlatr, seems like this branch doesnt have too many conflicts with main, do you want to rebase it so we can merge ?

Issue #116, opened 2021-10-30
Implement PublicPropertyForm validation


2ynn:
For PublicPropertyForm: https://github.com/LesFruitsDefendus/saskatoon-ng/blob/f6a2a8cdadbcdedd1463837f2017cebfc89a3c1b/saskatoon/harvest/forms.py#L325
- [x] most importantly check for invalid email
- [ ] eventually phone number (there are defined as simple CharFields and aren't used for any automatic task so far)
- [ ] eventually postal code (same)
- [ ]  evaluate if Blank=False should be set on neighborhood field

tristanlatr:
This library should help us: https://validators.readthedocs.io/en/latest/

2ynn:
thanks @tristanlatr, validators are already available with django form fields https://docs.djangoproject.com/en/4.0/ref/validators/
 they just need to be properly implemented, e.g. in https://github.com/LesFruitsDefendus/saskatoon-ng/blob/master/saskatoon/harvest/forms.py

donaldte:
> thanks @tristanlatr, validators are already available with django form fields https://docs.djangoproject.com/en/4.0/ref/validators/ they just need to be properly implemented, e.g. in https://github.com/LesFruitsDefendus/saskatoon-ng/blob/master/saskatoon/harvest/forms.py

Hi I have used this validators available in django to implement email and phone number validation inside the models Please I want you to check and tell if there are something to modify. the branches are under develop branch. thanks

tristanlatr:
Hello @donaldte,

Please open a pull request with your changes and I'll review them alongside other contributions tonight or tomorrow alongside other contributions. 

Thanks

yokwejuste:
> thanks @tristanlatr, validators are already available with django form fields https://docs.djangoproject.com/en/4.0/ref/validators/ they just need to be properly implemented, e.g. in https://github.com/LesFruitsDefendus/saskatoon-ng/blob/master/saskatoon/harvest/forms.py

Hey greetings here! Can I work on this issue? I mean can the issue be assigned to me?

2ynn:
To be clear on this issue. Do not alter any `models.py`, we are in the process of migrating databases and cannot afford doing migrations at the moment.
The form that needs to be validated more specifically is PublicPropertyForm
[...]

PR #182, opened 2022-04-04
Update phone fields from CharField to PhoneField

farhanmasud:
Partially addresses #116 

`pending_contact_phone` field on `harvest.models.Property` now uses `PhoneField` from [django-phone-field](https://github.com/VeryApt/django-phone-field/)

The `PhoneField` is basically a modified `CharField` and has the default the `max_length` argument set to `31` as default and for that I have removed the previously set `max_length=50`. Phone numbers probably doesn't have that many characters and I believe this might not create any conflicts (e.g. any previous instance having data in this field that has more length that 31) if existing stored data needs to be modified with new rules.

Question here is that do we need to modify existing data in this `pending_contact_phone` field to be saved with the validations of `PhoneField` as well? Because after running the new migrations, only new numbers will be saved with the `PhoneField` rules and the old ones will be left as is.

One other question, should I update this PR for other field (e.g. `pending_contact_name`, `pending_contact_family_name`, `postal_code`, etc) updates


2ynn:
@farhanmasud thanks for this. I've just tested and the old phones are indeed kept as is (not stored as E.164), but at least calling `str(property.pending_contact_phone)` formats them nicely. We do not need to clean all old properties' pending contact phones in the database. 
That being said, since the migration works nicely, I would suggest to do the same for `Person.phone` and `Organization.phone` in the member app. These ones though should eventually get cleaned up in our database (out of scope for this contribution).

Are you okay with adding these two other fields to this PR? We could then handle `postal_code` and splitting names in other PRs. In which case maybe let's call this PR something like `Update phone fields from CharField to PhoneField` ?

farhanmasud:
@2ynn thanks for the feedback. I'm updating the other two fields in this PR accordingly.

farhanmasud:
@2ynn I have updated `Person.phone` and `Organization.phone` with `PhoneField` as well along with the migrations. I have removed `max_length` from both of these as the `PhoneField` has a default `max_length=31` and to keep same with the `harvest.models.Property.pending_contact_phone` field.

2ynn:
thanks @farhanmasud, looks good to me, I'll test locally when I get a chance

farhanmasud:
@2ynn I have updated the function `get_similar_properties` in `harvest.utils` as well to use this field.

2ynn (approved):
thanks a lot @farhanmasud, I'll merge shortly

PR #185, opened 2022-04-04
[#34] Add support for email notification to tree owner

### LINKED ISSUE #34: Notify tree owner once "pending" status is removed
Send automatic email to tree owner to indicate that their tree subscription was reviewed and completed.
From https://github.com/tiagovaz/saskatoon/issues/192

victorphoenix3:
Adds support for email notification to tree owner when pending status is removed by administrator.

Fixes #34 
Depends on #213 

### DIFF
--- a/saskatoon/harvest/models.py
+++ b/saskatoon/harvest/models.py
@@ -774,6 +774,11 @@ models.signals.pre_save.connect(
     sender=Property
 )
 
+models.signals.pre_save.connect(
+    receiver=signals.notify_pending_status_update,
+    sender=Property
+)
+
 models.signals.post_save.connect(
     receiver=signals.clear_cache_property,
     sender=Property
--- a/saskatoon/harvest/signals.py
+++ b/saskatoon/harvest/signals.py
@@ -2,6 +2,7 @@ from crequest.middleware import CrequestMiddleware
 from django.core.mail import send_mail
 from django.core.cache import cache
 from saskatoon.settings import SEND_MAIL_FAIL_SILENTLY
+from django.utils.translation import gettext_lazy as _
 
 def clear_cache_property(sender, instance, **kwargs):
     cache.delete_pattern("*property*")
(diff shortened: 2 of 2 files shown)


saskatoon/harvest/signals.py
```diff
@@ -37,6 +37,29 @@ def changed_by(sender, instance, **kwargs):
...
+def notify_pending_status_update(sender, instance, **kwargs):
+    # Send email only if pending status is removed
+    if instance.id:
+        original_instance = sender.objects.get(id=instance.id)
+        if original_instance.pending and not instance.pending:
+            property_owner_email = list()
+            if instance.owner:
+                contact_person = instance.owner.get_person()
+                if not contact_person:
+                    contact_person = instance.owner.get_organization().contact_person
```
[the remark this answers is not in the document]
2ynn: yes ideally it should never happen that an `Actor` is neither a `Person` nor an `Organization` but we do have 3 "Unknown Actors" in the database right now. So yes I agree it'd be better to catch if `instance.owner.get_organization()` returns None.

All that being said @tristanlatr I think I have been very confused with this and I think I have implemented many things where I assume a property owner is a `Person` and I didn't realize an `Organization` could also have a property (we have 16 in the database right now). So I think I should open a separate issue for this and we should adapt the "register owner" workflow to the case where it is an organization.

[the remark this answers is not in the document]
tristanlatr: Sounds good

saskatoon/harvest/signals.py
```diff
@@ -37,6 +39,46 @@ def changed_by(sender, instance, **kwargs):
...
+            property_owner_email = list()
+            if instance.owner:
+                contact_person = instance.owner.get_person()
+                if not contact_person:
+                    try:
+                        contact_person = (
+                            instance.owner.get_organization().contact_person
+                        )
+                    except:
+                        raise RuntimeError("Property owner is neither a person nor an organization")
```
tristanlatr, line 56: There are three unknowns properties that will make this block execute If I understand correctly.
I’d suggest we simply return None instead of crashing.

tristanlatr (approved):
Hi @victorphoenix3, thanks a lot and sorry for the late reply.

I think it looks good and can be merged. Did you had a change to configure saskatoon email alerts on your local computer ?

I'll let @2ynn merge this PR, when you'r ready.

Talk to you later

victorphoenix3:
I will update this PR as per https://github.com/LesFruitsDefendus/saskatoon-ng/pull/213

saskatoon/harvest/models.py
```diff
@@ -774,6 +774,11 @@ def __str__(self):
     sender=Property
 )
 
+models.signals.pre_save.connect(
```
tristanlatr, line 777: Just a little question, why using `pre_save` and not `post_save` here? 
If there is an error while calling `save()`, I suppose the email would have been triggered anyway.

Tell me what you think,

2ynn: I think `pre_save` is used to detect the change in status (`original_instance.pending`). After saving the object you'd loose track of what the former value of `pending` used to be. An alternative to avoid relying on signals would have been to manually store `self.pending` in the model's `__init__` and compare it to the new value in the `save()` method but i find the `pre_save` pretty elegant.

tristanlatr: ok very well, this was a misunderstanding.

victorphoenix3:
@2ynn shouldn't this PR be merged into the branch `Fix-188-Not_all_owners_are_persons` since it depends on methods that were recently added to this branch and which are not yet integrated into the `develop` branch?

2ynn:
@victorphoenix3 Ugh my bad, you clearly indicated it in first comment. I just rebased `Fix-188-Not_all_owners_are_persons` onto develop to at least include your changes

PR #196, opened 2022-04-06
Update pull_request_template.md

mina-andajos:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -7,7 +7,7 @@ Fixes #issue-number
 ## Type of change:
 - [ ] Bug fix (non-breaking change which fixes an issue).
 - [ ] New feature (non-breaking change which adds functionality).
-- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
+- [ ] Changes to models (requires making migrations).
 - [ ] Documentation change.
 ----------
 ## What's Changed:
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,6 @@
 <!-- 
 Thanks for your contribution!
-Please fill out the neccessary sections, and delete unused ones.
+Please fill out the necessary sections, and delete unused ones.
 -->
 Fixes #issue-number
 ----------
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,3 +1,35 @@
 <!-- 
 Thanks for your contribution!
+Please fill out the neccessary sections, and delete unused ones.
 -->
+Fixes #issue-number
+----------
+## Type of change:
+- [ ] Bug fix (non-breaking change which fixes an issue).
+- [ ] New feature (non-breaking change which adds functionality).
+- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
+- [ ] Documentation change.
+----------
+## What's Changed:
+- one change.
+- another change.
+- and so on.
+
+--------
+## Screenshots:
+1. one screenshot.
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,3 +1,35 @@
 <!-- 
 Thanks for your contribution!
+Please fill out the neccessary sections, and delete unused ones.
 -->
+Fixes #issue-number
+----------
+## Type of change:
+- [ ] Bug fix (non-breaking change which fixes an issue).
+- [ ] New feature (non-breaking change which adds functionality).
```
tristanlatr, line 9: There is not concept of breaking change in saskatoon right now. So I don;t think we should add this stuff in perenthesis.

.github/pull_request_template.md
```diff
@@ -1,3 +1,35 @@
...
+- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
```
tristanlatr, line 10:
```suggestion
```

.github/pull_request_template.md
```diff
@@ -1,3 +1,35 @@
...
+
+-------
+## Checklist:
+- [ ] My code follows the [style guidelines](https://github.com/LesFruitsDefendus/saskatoon-ng/blob/develop/CONTRIBUTING.md) of this project.
+- [ ] I have performed a self-review of my own code.
+- [ ] I have commented my code, particularly in hard-to-understand areas.
+- [ ] I have made corresponding changes to the documentation.
+- [ ] My changes generate no new warnings.
+- [ ] New and existing unit tests pass locally with my changes.
+- [ ] Any dependent changes have been merged and published in downstream modules.
```
tristanlatr, line 35: I don't think this is relevant for us
```suggestion

```

tristanlatr (changes requested):
Hi @mina-andajos,

Thanks a lot, it looks great.

.github/pull_request_template.md
mina-andajos: Wouldn't for example dependency changes or some major changes break the environment for some people?
I could maybe add this instead:

```markdown
- [ ] Change to the app models (requires making migrations).
```
What do you think? @tristanlatr

tristanlatr: Yes that sounds good

.github/pull_request_template.md
```diff
@@ -1,3 +1,34 @@
 <!-- 
 Thanks for your contribution!
+Please fill out the neccessary sections, and delete unused ones.
 -->
+Fixes #issue-number
+----------
+## Type of change:
+- [ ] Bug fix (non-breaking change which fixes an issue).
+- [ ] New feature (non-breaking change which adds functionality).
```
tristanlatr, line 9:
```suggestion
- [ ] New feature (change which adds functionality).
```

.github/pull_request_template.md
```diff
@@ -1,3 +1,34 @@
 <!-- 
 Thanks for your contribution!
+Please fill out the neccessary sections, and delete unused ones.
 -->
+Fixes #issue-number
+----------
+## Type of change:
+- [ ] Bug fix (non-breaking change which fixes an issue).
```
tristanlatr, line 8:
```suggestion
- [ ] Bug fix (change which fixes an issue).
```

tristanlatr (changes requested):
Looks good, only that breaking change is not a thing for us at the moment, so I'de suggest simply not use that term since we can't define it correctly.

tristanlatr (approved):
Thanks @mina-andajos

Issue #165, opened 2022-04-02
On cloning, installing the requirements and running the Saskatoon-ng repo some errors occured


yokwejuste:
## Background
I don't really know how it came about, but recently my Ubuntu 20.04 crashed and the kernel was damaged, 2 days ago I reinstalled the OS but I preferred an upgrade to 21.04.
## Expectations
Obviously, I expected to run the as previously did before the crash but got hocked up.
## what happened
- Finally, I got a first error installing some requirements, mainly at the level of mysqlclient. Below is an illustration or snip of the said error.
![image](https://user-images.githubusercontent.com/71908316/161365492-d69c82ac-63f3-4eb1-b66e-b2736f695cc1.png)
- Then at the level of trying to start the project using mysqlite3, other issues occured.
![image](https://user-images.githubusercontent.com/71908316/161365535-b4470098-ba54-4d50-91f6-eb5d7585352a.png)
## Suggestion
Maybe it's lacking in my requirements or something else? but so far my *pip freeze* gave me these modules
![image](https://user-images.githubusercontent.com/71908316/161365650-f557a571-24ea-4e70-9596-0f2f6d2e9334.png)
![image](https://user-images.githubusercontent.com/71908316/161365630-6575a06b-cd77-4dca-81bc-1b557c11e51a.png)

Can I get any help here?

tristanlatr:
Hi @yokwejuste,

I think you are definitely missing some requirements.

Here is my pip freeze output. after `pip install .`:

```
(ve) tristan@Tristans-MBP saskatoon-ng % pip install .
(ve) tristan@Tristans-MBP saskatoon-ng % pip freeze
asgiref==3.3.4
attrs==20.3.0
certifi==2021.10.8
cffi==1.14.0
charset-normalizer==2.0.12
cryptography==3.4.7
Django==3.2
django-appconf==1.0.4
django-autocomplete-light==3.8.2
django-bootstrap-form==3.4
django-bootstrap3==15.0.0
django-ckeditor==6.0.0
django-crequest==2018.5.11
django-crispy-forms==1.11.2
django-debug-toolbar==3.2.4
django-dotenv==1.4.2
django-extensions==3.1.5
django-filter==2.4.0
django-geojson==3.0.0
django-js-asset==1.2.2
django-leaflet==0.26.0
django-redis==4.12.1
django-rest-knox==4.1.0
django-rosetta==0.9.8
django-select2==7.7.1
[...]

PR #203, opened 2022-04-07
Update INSTALL.md #165

richarddushime:
On cloning, installing the requirements and running the Saskatoon-ng repo some errors occured
#165
to solve this issue  i updated even  env.template file by adding the export keyword  the  variables

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -26,8 +26,11 @@ Please follow each part of this documentation in order to run your own instance 
     ```
 
     $ virtualenv venv
+    
     $ . venv/bin/activate
+    
     $(venv) pip3 install .
+    
     ```
 
     See `setup.py` for more details on the project's package requirements
@@ -38,6 +41,9 @@ Please follow each part of this documentation in order to run your own instance 
     `django-redis` is currently used as a caching backend with default configuration (see `CACHES` variable in `saskatoon/settings.py`). A Redis server must be run in the background:
     ```
     $ sudo apt install redis-server
+    ```
+    checking  status of redis-server
+    ```
     $ sudo systemctl status redis-server
     ```
     Note the Redis service will start automatically when the installation finishes (if using systemd)
@@ -79,6 +85,8 @@ Note: to generate a new random secret key, you can run this script:
 python -c 'from django.core.management.utils import get_random_secret_key; print(get_random_secret_key())'
 
 ```
+After Generating the Secret Key Copy and paste it in your ` .env ` file 
+
 
 ## local database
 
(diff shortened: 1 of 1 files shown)


tristanlatr (approved):
Hi @richarddushime,

Glad to see that your issue got resolved. 

I feel like this change should not be necessary for django-dotenv to read your config file.
But since [it supports adding the export keyword](https://github.com/jpadilla/django-dotenv/blob/c68fd4bf3ae330c7160a620210f760f0a52359fd/tests.py#L68), I don't see any reason why not adding it.

INSTALL.md
```diff
@@ -115,11 +123,11 @@ $ sudo mysql -u root -p
 Example mysql configuration in *.env* file:
```
2ynn, line 123: @richarddushime here we specify the content goes in the `.env` file (to be parsed by `django-dotenv`)

2ynn:
@richarddushime same i'm gonna close this one, let me know if you still have issues importing environment variables by adapting the current `.env` template file.

2ynn:
@richarddushime  @tristanlatr sorry just saw the comment about `django-dotenv` supporting the `export` keyword. My bad!
I don't think it's necessary though. Let me know how you feel about it.

richarddushime:
> @richarddushime @tristanlatr sorry just saw the comment about `django-dotenv` supporting the `export` keyword. My bad! I don't think it's necessary though. Let me know how you feel about it.

on my side i faced several errors while setting up the project and i tried to fixed the issue in several ways  still  could read  the .env  file variables  at the end i got a  quick solution for the problem  by using the ``` export ```keyword in the file though there is([supports adding the export keyword](https://github.com/jpadilla/django-dotenv/blob/c68fd4bf3ae330c7160a620210f760f0a52359fd/tests.py#L68),)  Actually i also guided someone  who was facing the same issue   and it worked well with export keyword in .env file

tristanlatr (changes requested):
Yeah, 

I'm not sure we should introduce these changes, I kinda agree with @2ynn  on that. 
I think the issue resides in the location of the `.env` file being for clear enough, but I'm not certain.

Sorry for the trouble setting up saskatoon @richarddushime.

The little docs additions in the `INSTALL.md` looks good nonetheless.

2ynn:
Yea I agree the sentence "Copy the `saskatoon/env.template` file into `saskatoon/.env` and adapt it to your needs." could be improved.
What I meant by this is that the `.env` file must be copied at the same location as the template file, i.e.:
```
cp saskatoon/env.template saskatoon/.env
```
And "to your needs" could for sure be improved, (e.g. detailing that the secret key must be added to the file after being generated etc...)

PR #204, opened 2022-04-07
updated env.template

richarddushime:
After updating the INSTALL.md   here  is env.template updated


tristanlatr (approved):
Looks good, thanks.

2ynn:
@richarddushime the `.env` file is intended to be read by `django-dotenv`, not be sourced in your shell. 
https://pypi.org/project/django-dotenv/

tristanlatr:
> .env file is intended to be read by django-dotenv, not be sourced in your shell.

Technically both should work, but I agree this could introduce misunderstandings about setting up saskatoon, Because we would have different manner to set the environment variables.

2ynn:
@richarddushime I'm closing this one, feel free to reach out if the environment variables setup is still not clear.

PR #215, opened 2022-04-09
[#202] Fix sticky footer

### LINKED ISSUE #202: change sticky-footer to fixed-footer, so it is always at the bottom
hello @tristanlatr i would like to make changes to the sticky footer so it always shows up at the bottom on large screens and all over the app. Can i work on this?
lesfruit2

nnabuihe-favour:
Fix #202


saskatoon/saskatoon/.gitignore
```diff
@@ -0,0 +1,4 @@
+.env
+/.env
+/saskatoon/.env
+saskatoon/.env
```
2ynn, line 4: we already have these in gitignore:
```
.env
.venv
env/
venv/
ENV/
env.bak/
venv.bak/
```
I believe that `.env` pattern should already ignore your `saskatoon/.env` file. Please confirm

saskatoon/saskatoon/settings.py
```diff
@@ -21,7 +21,7 @@
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
 # Load the environment variables from .env file.
-dotenv = os.path.join(BASE_DIR, '.env')
+dotenv = os.path.join(BASE_DIR, 'saskatoon/.env')
```
2ynn, line 24: BASE_DIR should be `saskatoon` here, please double check

nnabuihe-favour: i started the development server with python3 saskatoon/manage.py runserver 8000 as is the install.md. A nd i got this error
env

saskatoon/saskatoon/.gitignore
nnabuihe-favour: Yes i placed it in the wrong folder, that's why i had this issue. i've resolved mine locally thanks a lot

saskatoon/saskatoon/settings.py
nnabuihe-favour: so i manually had to set 'saskatoon/.env' which fixes the problem. I guess it ought to be in the saskatoon folder not the saskatoon/saskatoon folder.

2ynn: where did you place your `.env` file exactly? If you place it at the same location than the template file it should work since:
- `BASE_DIR` should give `../saskatoon/`
- `dotenv` should be `../saskatoon/.env` as indicated in the INSTALL.md, not `saskatoon/saskatton/.env`

PR #217, opened 2022-04-09
fix all missing translations

donaldte:
Hi @tristanlatr  @2ynn 
I have gone through the project and corrected all the missing translations. I also noticed that the HTML files for the statistics are still under development so I did not modify anything there.
After it's modification if you have to update the .po and .mo file
I did it locally but I didn't make a commit for it for fear that it would create problems so if you find that the various modifications I made are necessary after having merged you will have to execute the following commands :
django-admin makemessages -l fr
django-admin makemessages -d djangojs -l fr  ..this  for javascript file if there is one
django-admin messagecompiler

Thanks


tristanlatr (approved):
Hello @donaldte, thanks a lot for all these translations. I've tested it locally and works good.

Issue #117, opened 2021-10-30
Add editable HomePage model


2ynn:
To allow core-members to dynamically edit the content of the home page:
- title (Welcome)
- image (Banner)
- description (LFD about)
- volunteer instructions (rendered as |safe)
- pickleader instructions (rendered as |safe)

See also FAQ in issue #29

khabdrick:
is this issue still available, I'll like to work on it?

tristanlatr:
Hello @khabdrick,

Note that this issue was initially part of the internship itself, because it would probably introduce quite a big diff.

But if you feel confident you can achieve task in the contribution period, please go ahead.

khabdrick:
Now I don't feel confident anymore. I'll pick something lighter so that it can be finished and reviewed in time.
I'm sorry for the confusion.

2ynn:
no worries, sorry for the django-dotenv / python-dotenv confusion, that was on our side

tristanlatr:
@khabdrick you can review the issues marked as "easy" or "good first issue" :)

Talk to you later,

2ynn:
no worries @khabdrick !

victorphoenix3:
Hi, I would like to work on this issue?

What I am thinking is that instead of hard coding the values in templates, we should introduce a model with the fields that can be edited in the admin panel and then we could render the fields dynamically. Do you think this would work? @2ynn @tristanlatr

tristanlatr:
> we should introduce a model with the fields that can be edited in the admin panel and then we could render the fields dynamically. 

yes this is the plan :) as I mentioned above, please make sure you feel confident to complete this task within the contribution period before working in this.

Talk to you later,

PR #218, opened 2022-04-10
[#117] Make pickleader instructions editable

victorphoenix3:
- [x] pickleader instructions (rendered as |safe)

Fixes: #117


saskatoon/sitebase/templates/app/index.html
```diff
@@ -22,7 +22,11 @@ <h2>{% trans "Hi" %} {{user}} !</h2>
                                         <hr>
                                         {% include 'app/index/pickleader.html' %}
                                     {% else %}
-                                        <h2>{% trans "Welcome to Les Fruits Défendus!" %}</h2>
+                                        {% if homepage %}
```
victorphoenix3, line 25: In case the user does not create entry in the homepage model, we should display the dafault welcome message.

victorphoenix3:
@tristanlatr Hi, can you please clarify where the description and banner image is?

tristanlatr:
Hi @victorphoenix3,

Thanks to work on this, I think that for the first PR that introduces an editable home page, it's ok if there is not image support yet. I can add more complexity at a time where I'm not sure it would be helpful. So I'de suggest to forget about images for now. 

I think the most important thing here is to have an editable body as `markdown`. To convert markdown to HTML with few bonuses I'd suggest the following approach:

```python
def md_to_html(text:str) -> str:
  return markdown.markdown(text, 
        extensions=['pymdownx.highlight', 'pymdownx.superfences', 
                    'pymdownx.details', 'pymdownx.magiclink', 
                    'markdown.extensions.toc', 'markdown.extensions.admonition'])
```
This is the actual content that should be easily editable, the rest can stay static.
by default 2022-04-10 at 10 30 56 PM

So there is only the markdown body field that is really important in my opinion. Content that changes depending on current role of the login person (if any).

Sorry if the ticket wasn't clear about the main goal: which is simply not to have the information above hardcoded.

victorphoenix3:
@tristanlatr  I have made use of `Django Markdownify`. Its a wrapper over [Markdown](https://pypi.python.org/pypi/Markdown) and [Bleach](http://pythonhosted.org/bleach/index.html) and lets us use a simple tag that converts `html` to `markdown` and uses `safetext`.

I am not sure how to manage translations here. So I have created two models, one for French and other English.

saskatoon/sitebase/models.py
```diff
@@ -0,0 +1,17 @@
...
+
+class Homepage(models.Model):
+
+    pickleader_instructions = models.TextField(
+        blank=True,
+        null=True,
+        verbose_name=_("Pickleader Instructions in English"),
+        help_text=_("Steps to pick a leader"),
+        max_length=1000
+    )
```
2ynn, line 12: I believe we need to either have both english and french versions for the TextField, or have a language field, e.g.
    - volunteer_instructions
    - pickleader_instructions
    - language (EN/FR/..)
    - status (ACTIVE / DRAFT)

2ynn (commented):
Let's not rush into this one, we need to better specify the requirements in #117

2ynn:
implemented in #383

PR #223, opened 2022-04-11
Fixed issue #202

nnabuihe-favour:
Appropriate fix for issue #202 without env conflicts. please neglect the previous commit


saskatoon/sitebase/static/css/bootstrap4.min.css
```diff
@@ -0,0 +1,7 @@
+/*!
+ * Bootstrap v4.0.0 (https://getbootstrap.com)
```
tristanlatr, line 2: i think we're already using bootstrap v3 somewhere, I' mo unsure if we should use both...

tristanlatr (changes requested):
Is is possible to make theses changes without included bootstrap4 as a whole ? 
If we're using bootstrap4, I think we should drop bootstrap3, having both could introduce conflicts.

saskatoon/sitebase/static/css/bootstrap4.min.css
nnabuihe-favour: yes the bootstrap v4 is just for the crispy form. the rest of the codebase still uses v3

nnabuihe-favour: About the issue of conflicts, i think it is possible to work on it via the css tags used by bootstrap which are <b>,<strong>but manually adding a fix to these tags will affect other parts of the code base that use the tag. And trying to use classes would mean trying to edit it from the crispy end. i'll work on a quick fix and try to make it stable.

tristanlatr: So that's a good reason not wanting to add both bootstrap 3 and 4, I think.

tristanlatr:
Hello @nnabuihe-favour, sorry about this but looks like someone has worked on your issue, see #254.
Since I don't want to introduce the whole bootstrap 4 while keeping the bootstrap 3, I'm going to close this PR.

Thanks for your understanding,

Issue #222, opened 2022-04-10
Fix PersonCreateForm validation


2ynn:
`forms.ValidationError` does not show in `PersonCreateForm` when trying to create a new user with duplicate email address

To reproduce, 
1. create a new user using an email address that is already taken in the database:
http://localhost:8000/person/create/
2. form does not save but no error message is prompted
3. change the email for a new email address. the form saves properly

farhanmasud:
I'm working on the `PublicPropertyForm` validation on [#116](https://github.com/LesFruitsDefendus/saskatoon-ng/issues/116), I'd like to work on this issue since both are related to form validations.

farhanmasud:
I think I have found the bug. I'm making a PR shortly.

Issue #221, opened 2022-04-10
Fix register owner based on pending property


2ynn:
in `/person/create/?pid=<pending_property_id>/`

https://github.com/LesFruitsDefendus/saskatoon-ng/blob/f032e5a9a5b717df43acc4f3ef8e1daed5350016/saskatoon/member/views.py#L26

now breaks because phone is no longer a string but a PhoneField: `'PhoneNumber' object has no attribute 'replace'`

To be tested:
```
 'phone': p.pending_contact_phone
```

Note: this functionality assumes the owner is a Person, not an Organization. We should maybe consider adding a `is_organization` checkbox to the public property form

farhanmasud:
@2ynn I'd like to work on this if you didn't already spend time working on a fix for this (asking as if you've self assigned this). As I worked with the codebase for making the changes with the `PhoneField` I think I can find a fix for this.

2ynn:
@farhanmasud ok yes please go ahead, I quickly tested and `'phone': p.pending_contact_phone.` will work just fine. You can add the change along with your other PR for issue #222 if you want.

farhanmasud:
> Note: this functionality assumes the owner is a Person, not an Organization. We should maybe consider adding a `is_organization` checkbox to the public property form

@2ynn what will be the backend logic for this field? I mean, where we should store the value that user inputs here? Does a new model field need to be created? Or should this value programmatically alter value for some other field (e.g. save a new instance of `Organization` model)?

There is already a property is being developed with the same name `is_organization` as per [#188](https://github.com/LesFruitsDefendus/saskatoon-ng/issues/188) that you have suggested. Creating a new model field with the same name will conflict the property name when they are merged.

2ynn:
@farhanmasud let's leave that aside for now. There are not a lot of organizations that own trees (most are beneficiary organizations) and they can be contacted separately by core members (without using the public form). So it's okay to assume all pending properties are for `Person` owners for now.
That being said I would like to add a safeguard in `PersonCreateForm`: https://github.com/LesFruitsDefendus/saskatoon-ng/blob/ebc6660b5c996463245472eb294161df38cc33bd/saskatoon/member/forms.py#L56
```
        pid = self.cleaned_data['pending_property_id']
        if pid:
            try:
                pending_property = Property.objects.get(id=pid)
                pending_property.owner = instance
                pending_property.save()
            except Exception as e: print(e)
```
[...]

PR #224, opened 2022-04-11
[#222] Fix PersonCreateForm validation

farhanmasud:
Fixes [#222](https://github.com/LesFruitsDefendus/saskatoon-ng/issues/222)

After submitting the form with an email already existing in the database, there was no `POST` request showing up on the console (`python saskatoon/manage.py runserver`). The issue was probably when the form was submitted and it was invalid, it wasn't handled properly. To fix this I have used `form_invalid` method in `PersonCreateView` to handle the invalid form.

After this there is the `POST` request showing up on the console after submitting the form with an email that already exists in the database -

Screenshot from 2022-04-11 13-00-21 png-mh

But this change still wasn't showing the errors. To fix this, I have changed `clean_email` method under `PersonCreateForm` to the default `clean` method and updated as needed. Now it shows the errors on the page.

Screenshot from 2022-04-11 12-56-24

Also work in progress for issue [#221](https://github.com/LesFruitsDefendus/saskatoon-ng/issues/221)

### DIFF
--- a/saskatoon/member/forms.py
+++ b/saskatoon/member/forms.py
@@ -40,7 +40,7 @@ class PersonCreateForm(forms.ModelForm):
     field_order = ['roles', 'first_name', 'family_name', 'email', 'language']
 
     def clean(self):
-        cleaned_data = super(PersonCreateForm, self).clean()
+        cleaned_data = super().clean()
         validate_email(cleaned_data['email'])
 
     def save(self):
@@ -52,11 +52,12 @@ class PersonCreateForm(forms.ModelForm):
                 email=self.cleaned_data['email'],
                 person=instance
         )
-        auth_user.set_roles(self.cleaned_data['roles'])
+        roles = self.cleaned_data['roles']
+        auth_user.set_roles(roles)
 
         # associate pending_property (if any)
         pid = self.cleaned_data['pending_property_id']
-        if pid:
+        if pid and 'owner' in roles:
             try:
                 pending_property = Property.objects.get(id=pid)
                 pending_property.owner = instance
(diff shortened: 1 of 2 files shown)


saskatoon/sitebase/templates/app/forms/property_create_public.html
```diff
@@ -15,6 +15,7 @@
                     <div class="form-example-wrap">
                         <div class="cmp-tb-hd">
                             <h2>Add a new property</h2>
+                            <small class="text-muted">This form is for persons with a tree on their property and community organizations should contact us directly</small>
```
2ynn, line 18: I'm not sure if this notice is really necessary here. It's okay if we receive pending properties from organizations as long as we have some kind of way of knowing (e.g. core members give specific instructions on how to add a note in the name or additional info box). Btw we could already edit the `help_text` in the `additional_info` box to hint at it.
This way core members could tick the `Contact Person` instead of `Property owner` when they use the register owner form. 
(They will then need to manually create an Organization, link it to the contact person, and finally manually add the organization as Owner). We will make sure to clarify this flow in our guides with core members.

Also I'm wondering if we could improve the title for something like <h2>Register a property for the {current_year} season</h2>
Some properties aren't "new" since they can check the "have you registered" before box.
@tristanlatr what do you think? We could also open a separate issue to finalize the look/labels/translations in the public form

farhanmasud: @2ynn got it, thanks for elaborating the process of how organization creation works.

I'm removing the last commit for now that deals with this help text. Other than this, I think my other changes are ready to be reviewed.

We can work on the look/labels/translations related changes on a different issue as you have suggested.

tristanlatr:
I'll let @2ynn merge this PR.

2ynn (approved):
Thanks @farhanmasud looks great!!!

PR #230, opened 2022-04-11
Registration login system

Codde24:
I have slightly altered login system. i have enabled a user to register for an account in this system with an active status of false. The admin receives a email notification of this application and can then approve the application. When the application is approved the user receives an email notification and can then login to his account

### DIFF
--- a/saskatoon/member/models.py
+++ b/saskatoon/member/models.py
@@ -7,6 +7,11 @@ from django.contrib.auth.models import ( Group, AbstractBaseUser,
                                          PermissionsMixin, BaseUserManager )
 from django.core.validators import RegexValidator
 from harvest.models import RequestForParticipation, Harvest, Property
+from django.dispatch import receiver
+from django.db.models.signals import pre_save, post_save
+from django.conf import settings
+from django.core.mail import send_mail
+
 
 AUTH_GROUPS = (
     ('core', _("Core Member")),
@@ -26,7 +31,7 @@ class AuthUserManager(BaseUserManager):
 
         user = self.model(email=self.normalize_email(email),
                           )
-        user.is_active = True
+        user.is_active = False
         user.set_password(password)
         user.save(using=self._db)
         return user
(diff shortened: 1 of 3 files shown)


2ynn:
thanks @Codde24  for the suggestion!
@tristanlatr does this fit with the core members workflow though? I feel like users (i.e. pickleaders) should still be created manually by admins and registration should not be made available to anyone.

tristanlatr:
Hi @Codde24, thanks for this. it shall easy the annual registration of new pick leaders. 

I'm currently uncertain about the implementation, I'll need to look into it closer and try it out locally. 

But for sure, I think a global switch (in the admin panel) to enable/disable registration would be good. Because we would use this feature only a couple of weeks per year. 

And it does not covers the authorizations groups, an admin must edits those groups afterwards anyway, right ?
Maybe we could let the participant select how they got to this page and depending on the answer, we'd suggest pickleader or picker role. 

Talk to you later,

2ynn:
@tristanlatr Yeah I've already simplified the registration of new users by core members using the PersonCreateForm (which covers auth. groups assignment trough the roles checkbox list) in addition to the `reset-password` button in the Community panel. I've shown it to Simone and it seemed to be straightforward enough.

Codde24:
> Hi @Codde24, thanks for this. it shall easy the annual registration of new pick leaders.
> 
> I'm currently uncertain about the implementation, I'll need to look into it closer and try it out locally.
> 
> But for sure, I think a global switch (in the admin panel) to enable/disable registration would be good. Because we would use this feature only a couple of weeks per year.
> 
> And it does not covers the authorizations groups, an admin must edits those groups afterwards anyway, right ? Maybe we could let the participant select how they got to this page and depending on the answer, we'd suggest pickleader or picker role.
> 
> Talk to you later,

@tristanlatr Ok sir thanks for this feedback..and yes an admin can still edit the authorization groups after

2ynn:
sorry @Codde24 our onboarding process has been updated using a different approach

PR #232, opened 2022-04-11
[WIP] Add comments box to properties

### LINKED ISSUE #21: Add comments box to properties
Allow users to write comments to properties

mina-andajos:
Fixes #21
----------
## Type of change:
- [ ] Bug fix (non-breaking change which fixes an issue).
- [x] New feature (non-breaking change which adds functionality).
- [ ] Documentation change.


mina-andajos:
@tristanlatr @2ynn 

What do you think of making an abstract comment model and then making comment models for harvests and properties (as in 076db30e0d68025bdb865c82c1adc952460cd936)?
I could add a foreign key for properties in the comment model instead but I believe that's bad for normalization.

mina-andajos:
> @mina-andajos sorry I just looked into issue #21 now. Why don't we use the `additional_info` field that we already have in `Property` model? @tristanlatr do you see an issue with pick-leaders adding comments into this field?

@2ynn 
Do you want to have a comment box like the harvest detail view?
Showing the author and time of comment.
image

tristanlatr:
Hello both of you,

Yes the idea was to have a list of comments like for harvests, but it's probably not a priority.

And also, even if I created the initial issue, I'm currently unclear about the usefulness of such feature.

So I'd suggest we work on other issues for now.

PR #234, opened 2022-04-12
Do not assume all owners are Persons

### LINKED ISSUE #188: Do not assume all owners are Persons
I've assumed all property owners were `Person`s without realizing `Organization`s can also be owners (`owner` field is related to `Actor`, not `Person`)
```
./saskatoon/sitebase/templates/app/list_views/property/status.html:21:    {% if property.owner.person.about %}
./saskatoon/sitebase/templates/app/list_views/property/status.html:22:        <li><i class="fa fa-envelope"></i> {{ property.owner.person.email }}</li>
./saskatoon/sitebase/templates/app/detail_views/property/similar_row.html:23:        {% with property.owner.person as po %}
./saskatoon/sitebase/templates/app/detail_views/harvest/about.html:53:                                <h4>{{ property.owner.person.phone }}</h4>
./saskatoon/sitebase/templates/app/detail_views/harvest/about.html:60:                                <h4>{{ property.owner.person.email }}</h4>
./saskatoon/sitebase/templates/app/detail_views/harvest/about.html:64:                        {% if property.owner.person.language != 'None' %}
./saskatoon/sitebase/templates/app/detail_views/harvest/about.html:68:                                <h4>{{ property.owner.person.language }}</h4>
./saskatoon/sitebase/templates/app/detail_views/harvest/about.html:73:                        {% if property.owner.person.comments %}
./saskatoon/sitebase/templates/app/detail_views/harvest/about.html:77:                                <h4>{{ property.owner.person.comments }}</h4>
./saskatoon/harvest/admin.py:99:        'owner__person__family_name',
./saskatoon/harvest/admin.py:100:        'owner__person__auth_user__email',
./saskatoon/harvest/utils.py:29:            query |= Q(owner__person__auth_user__email=email)
./saskatoon/harvest/autocomplete.py:26:            q2 = Q(owner__person__first_name__icontains=self.q)
./saskatoon/harvest/autocomplete.py:27:            q3 = Q(owner__person__family_name__icontains=self.q)
```
There may be more. See `app/list_views/equipment/data_row.html` for an example where a distinction between the two types of actors is made.

I suggest starting by adding "friendly" `@property` to the `Actor` model:
- `is_person`: returns True if the actor is a person (i.e. `hasattr(actor, 'person')`)
- `is_organization`: returns True if the actor is an organization (i.e. `hasattr(actor, 'organization')`)

Then to the `Property` model:
- `get_owner_name()`: person's name or organization's contact_person's name depending on `owner.is_person`/`owner.is_organization`
- `get_owner_email()`: person's email or organization's contact_person's email depending on `owner.is_person`/`owner.is_organization`
- `get_owner_phone()`: person's phone or organization's phone depending on actor type

Also I see this block commented out in `harvest.models`, I think it had a similar intent.
```
    # def get_owner_subclass(self):
    #     from member.models import Person, Organization
    #     try:
    #         return Person(self.owner)
    #     except Person.DoesNotExist:
    #         return Organization(self.owner)
```

2ynn:
Fix #188


2ynn:
@victorphoenix3 do you maybe have time to look at the diffs before I merge?

saskatoon/harvest/admin.py
```diff
@@ -69,25 +71,23 @@ class PropertyAdmin(LeafletGeoAdmin):
...
     list_display = (
-        '__str__',
-        'authorized',
+        'short_address',
+        'owner_edit',
+        'owner_type',
+        'owner_phone',
+        'owner_email',
         'pending',
-        'harvest_every_year',
```
tristanlatr, line 75: What happened to this field?

saskatoon/harvest/admin.py
```diff
@@ -69,25 +71,23 @@ class PropertyAdmin(LeafletGeoAdmin):
...
+        'owner_email',
         'pending',
-        'harvest_every_year', 
+        'authorized',
         'approximative_maturity_date',
-        'avg_nb_required_pickers',
         'neighborhood',
         'city',
         'postal_code',
+        'pk'
```
tristanlatr, line 85: What is the data that this field represent? 'pk' is not very verbose...

saskatoon/harvest/admin.py
```diff
@@ -69,25 +71,23 @@ class PropertyAdmin(LeafletGeoAdmin):
...
     list_filter = (
+        PropertyOwnerTypeAdminFilter,
         'authorized',
         'pending',
-        'harvest_every_year',
-        'public_access',
-        'neighbor_access',
-        'compost_bin',
-        'ladder_available',
-        'ladder_available_for_outside_picks',
```
tristanlatr, line 90: What happened to those fields?

tristanlatr (commented):
Few questions

saskatoon/harvest/admin.py
2ynn: I find having too many fields (especially with very long verbose_names) in the admin list_display makes the panel annoying to use due to the long horizontal scroll bar. Most of these fields can still be used as filters, I don't think they are primordial here personally.

saskatoon/harvest/admin.py
2ynn: pk is the Primary Key

saskatoon/harvest/admin.py
2ynn: sorry my previous comment had to do with removing list_display fields, in this snippet I did remove filters so we don't have to scroll *vertically* for too long. I think such filters will be mostly used by core members/pickleaders through the `Properties` web view and are not super important for admins but i might be wrong (in the old generation there were no fields or filters at all)

tristanlatr (approved):
Let’s do this

2ynn:
@tristanlatr after I hardcoded the season filter choices all checks have passed in CI/CD, including loading the `saskatoon.json` fixture.

https://github.com/LesFruitsDefendus/saskatoon-ng/pull/234/commits/4f7639a4977dcfe44e2761438b2163ce0dae95a3

tristanlatr:
> after I hardcoded the season filter choices

this doesn’t sounds very good…

2ynn:
well actually fetching Harvest objects within the `FilterSet` is not something we should do. I don't think setting the choices for season field dynamically is as straightforward as what was being done and this is what was causing the issue when importing `harvest.filters`
The only drawback is we would need to update this each year:
``
SEASON_FILTER_RANGE = (2016, 2022)
````

2ynn:
If really needed I can look about the proper way to override choices dynamically within the `FilterSet.__init__()` method but the way it was done was not correct.
here is a similar example:
https://stackoverflow.com/questions/55496921/reload-choices-dynamically-when-using-multiplechoicefilter

PR #238, opened 2022-04-13
Fix-Issue#6-Added help text hovering over property status

ejangphoebe00:
@tristanlatr implemented as requested, also which other issues should i take on after this.


2ynn (changes requested):
@ejangphoebe00 thanks for your contribution and I'm terribly sorry for the delay. Would you still be into working on this issue? If so I think we should also wrap all the new help texts you added with translation tags, i.e  `{% trans "...."  %}`

2ynn:
obsolete following #443

PR #272, opened 2022-04-17
Fixes  #172 without bootstrap 4

### LINKED ISSUE #172: Form validation styling issue with django-crispy-forms and bootstrap3 / bootstrap4 conflict
This front-end HTML template [currently uses bootstrap v3](https://github.com/LesFruitsDefendus/saskatoon-ng/blob/develop/saskatoon/sitebase/static/css/bootstrap.min.css) and in the settings file, the `CRISPY_TEMPLATE_PACK` variable is [currently set to `'bootstrap4'`](https://github.com/LesFruitsDefendus/saskatoon-ng/blob/62bc3b19e0654402bf92af99189497f6e6a1d168/saskatoon/saskatoon/settings.py#L228).

`bootstrap4` template pack of django-crispy-forms uses the CSS class `invalid-feedback` for rendering error messages under a field which doesn't exist in bootstrap v3. Bootstrap v3 uses the CSS class `help-block` for rendering the error messages instead.

As the templates use bootstrap v3, when django-crispy-forms with `'bootstrap4'` template pack is rendering error message with `invalid-feedback` CSS class, it's not displaying with the CSS styling required to show the error messages [for example, red color for font].

This came up while implementing the validations for phone field with as mentioned on [#116](https://github.com/LesFruitsDefendus/saskatoon-ng/issues/116) with [django-phone-field](https://pypi.org/project/django-phone-field/) on the `Property` model. Example screenshot provided below -

Screenshot from 2022-04-04 01-58-22

Here the error message `Only E164 numbers are supported here (+12223334444).` shows up in bold black text which was supposed to be in CSS classes to show errors in red font. Here the bootstrap4 class applied is `invalid-feedback` but since bootstrap3 doesn't come with this class, it's not showing up.

If I change the `CRISPY_TEMPLATE_PACK` variable to `bootstrap3`, it breaks the form, it's unable to render the two fields correctly that are in [`PhoneField`](https://github.com/VeryApt/django-phone-field/#usage). This particular case might be related to a different issue but I'm noting it here to demonstrate this particular scenario.

Screenshot from 2022-04-04 02-02-56

nnabuihe-favour:
changed crispy_template_pack to use bootstrap 3 and changed the css classes on the contact phone number and ext model.
Here's what it looks like before
contact

And after
contact_after

Thereby completely resolving #172 conflicts of css styling issue for both error font and margin while still retaining the codebase use of bootstrap 3

### DIFF
--- a/saskatoon/harvest/forms.py
+++ b/saskatoon/harvest/forms.py
@@ -478,6 +478,8 @@ class PublicPropertyForm(forms.ModelForm):
         self.fields['fruits_height'].widget.attrs['min'] = 1
         self.fields['street_number'].widget.attrs['min'] = 0.0
         self.fields['complement'].widget.attrs['min'] = 0.0
+        self.fields['pending_contact_phone'].widget.attrs.update({'class':'form-control'})
+
 
     def clean(self):
         cleaned_data = super(PublicPropertyForm, self).clean()
--- a/saskatoon/saskatoon/settings.py
+++ b/saskatoon/saskatoon/settings.py
@@ -226,7 +226,7 @@ CACHES = {
     }
 }
 
-CRISPY_TEMPLATE_PACK = 'bootstrap4'
+CRISPY_TEMPLATE_PACK = 'bootstrap3'
 
 CKEDITOR_CONFIGS = {
     'default': {


2ynn (commented):
Thank you so much @nnabuihe-favour for looking into this issue. 
changing to `bootstrap3` seems to break other forms. e.g. i can no longer check any box from the  `roles` MultipleChoiceField  in `PersonCreateForm`:
https://github.com/LesFruitsDefendus/saskatoon-ng/blob/96bfa793a46493cd287226271551e826e3424c74/saskatoon/member/forms.py#L28

Would you have an idea how to prevent this to happen? Thanks

Superpat:
@nnabuihe-favour Are you still interested on working on this ? If so you can rebase on main and we can do further testing.

PR #286, opened 2022-05-11
Fix 266 Add authorized=null option in PropertyFilter

### LINKED ISSUE #266: Property filters should match status name
Currently it show "Yes", "No" and "Unknown". Where we should read, depending on the status field, each statuses names + "All". 

So for instance for the authorization status: "Authorized", "Not Yet authorized", "Not authorized" and "All".

2ynn:
Fixes #266
----------
## Type of change:
- Bug fix (change which fixes an issue).
----------
## What's Changed:
- changed `authorized` BooleanFilter into ChoiceFilter in PropertyFilter to allow filtering by all three options: null / true / false
- current year is no longer hardcoded in HarvestFilter
- general cleanup in filters
- removed CRSF token in all filter forms (GET request)
----------
## Affected urls:
- localhost:8000/property/?authorized=2
- localhost:8000/property/?authorized=1
- localhost:8000/property/?authorized=0
-------
## Checklist:

### DIFF
--- a/saskatoon/harvest/filters.py
+++ b/saskatoon/harvest/filters.py
@@ -14,6 +14,16 @@ SEASON_FILTER_RANGE = (2016, datetime.now().year)
 
 class HarvestFilter(filters.FilterSet):
 
+    class Meta:
+        model = Harvest
+        fields = [
+            'status',
+            'pick_leader',
+            'trees',
+            'property__neighborhood',
+            'season',
+        ]
+
     YEARS = list(range(SEASON_FILTER_RANGE[0], SEASON_FILTER_RANGE[1]+1))
 
     season = filters.ChoiceFilter(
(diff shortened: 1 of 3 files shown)


saskatoon/sitebase/templates/app/list_views/filters.html
```diff
@@ -18,7 +18,6 @@ <h4 class="panel-title">
             <div class="collapse" id="accordionGreen-one" role="tabpanel">
                 <div class="panel-body" style="display:inline-block;">
                     <form role="form">
-                        {% csrf_token %}
```
tristanlatr, line 21: Why this change exactly ?

saskatoon/harvest/filters.py
```diff
@@ -50,51 +61,82 @@ class HarvestFilter(filters.FilterSet):
...
+
+    pending = filters.BooleanFilter(
+        help_text="",
+        label=_("Pending validation")
+    )
+
+    def authorized_filter(self, queryset, name, choice):
+        if choice == '2':
+            return queryset.filter(authorized__isnull=True)
+        return queryset.filter(authorized=bool(int(choice)))
```
tristanlatr, line 125: Would it make send to wrap this logic inside a custom `BooleanOptionalFilter` class? Maybe there are other optional boolean fields that would benefit from using it ?

This is a minor comment and does not have to be implemented in this PR.

tristanlatr (approved):
Looks great, thank you @2ynn

saskatoon/sitebase/templates/app/list_views/filters.html
2ynn: As far as I know a crsf token is not needed for a GET request and it's really annoying to have it in the url while testing filters

Issue #15, opened 2020-05-17
Create view for picker history/profile


tristanlatr:
See: https://github.com/tiagovaz/saskatoon/issues/75

tristanlatr:
This view could contains information about the user groups, have a link to change the password there instead of from the main menu.

- Have the possibility to edit it's Person / Organization (+ contact person) / Property record (we should not allow to edit the address of the property, we should invite people to create a new property with the public form).

- Have a list of all the harvest that the logged-in person has led ( good for ego ;) ).

PR #291, opened 2022-05-13
Fix #125 don't delete Organization on contact Person deletion

### LINKED ISSUE #125: Deleting a person linked to an organization deletes the org as well
The workaround is to associate the org with a another contact person before deleting the person.

But still, I feel like it should not delete the organization expect explicitly asked. 

by default 2022-03-24 at 12 45 51 PM

2ynn:
Fixes #125 
----------
## Type of change:
- Bug fix (change which fixes an issue).
----------
## What's Changed:
- set org's contact_person to null when deleting Person

-------
## Checklist:


tristanlatr:
Thanks @2ynn , would it be possible that  the form simply refuse to delete a person that is still linked to an organization. 
Indicating that another contact person should be chosen for the organization first. 

That would avoid further database inconsistencies,

Tell me what you think

2ynn:
> That would avoid further database inconsistencies,

We don't have a form to delete a Person, the garbage bin is still a link to the admin panel for now. We can clearly see which objects are deleted. Having `contact_person` set as a CASCADE at least let you realize it was still linked to the org, but you had to know what you were doing.

 As it is now an Organization doesn't have to have a contact_person (`null=True`) so setting it to null is valid.
 I think it boils down to know what you're doing,  I think if we move forward with a proper "Person" page (see Issue #15)  we'll be able to clarify all this and clearly indicate if the person is dangling or is still associated with an Org.

tristanlatr (approved):
Looks great

PR #296, opened 2022-05-21
Fix 292 Help cleanup Actor/AuthUser/Person inconsistencies

### LINKED ISSUE #292: Help clean up Actor/AuthUser/Person inconsistencies in prod database
Many contact persons aren't associated with any authuser, meaning they have no email address set.
TODO
- [x] add link to authuser in Person admin panel list display
- [x] add admin action in Person admin to create/associate person with authuser using dummy email
- [x] add filter to identify Persons with no AuthUser in PersonAdmin
- [x] improve Organization admin page to display contact person
- [x] add 'add to contact' admin action in AuthUser admin page
- [x] improve Actor admin page (display associated AuthUser, Person / Organization) 
- [x] Community tab: if person is a contact, show assiciated Organization in a column

2ynn:
Fixes #292
----------
## Type of change:
- [ ] Documentation change.
----------
## What's Changed:
See checklist in #292

--------
## Affected URLs:
http://localhost:8000/admin/member/actor/
http://localhost:8000/admin/member/organization/
http://localhost:8000/admin/member/authuser/
http://localhost:8000/admin/member/person/

-------
## Checklist:

### DIFF
--- a/saskatoon/member/admin.py
+++ b/saskatoon/member/admin.py
@@ -279,6 +279,18 @@ class PersonAdmin(admin.ModelAdmin):
         )
         return queryset
 
+    @admin.action(description="Create missing user(s)")
+    def create_user(self, request, queryset):
+        """Associate Person with new AuthUser object"""
+        nb_users = 0
+        for person in queryset.filter(auth_user__isnull=True):
+            email = f"FAKE@person.{person.pk}"
+            AuthUser.objects.create(email=email, person=person)
+            nb_users += 1
+        messages.add_message(request, messages.SUCCESS,
+                             f"Successfully created {nb_users} new users!")
+
+    actions = [create_user]
 
 
 @admin.register(Actor)
--- a/saskatoon/member/admin.py
+++ b/saskatoon/member/admin.py
@@ -6,6 +6,7 @@ from django.contrib.auth.forms import (UserCreationForm, UserChangeForm,
                                         ReadOnlyPasswordHashField)
 from django.db.models import Value
 from django.db.models.functions import Replace
+from django.urls import reverse
 from django.utils.html import mark_safe
 from member.models import (AuthUser, Actor, Language, Person, Organization,
                            Neighborhood, City, State, Country)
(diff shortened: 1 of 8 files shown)


saskatoon/member/admin.py
```diff
@@ -259,13 +270,35 @@ class PersonAdmin(admin.ModelAdmin):
         'auth_user__email',
     )
 
+    @admin.display(description="AuthUser")
+    def authuser(self, person):
+        try:
+            user = person.auth_user
+        except AuthUser.DoesNotExist:
+            return None
+        base_url = reverse('admin:member_authuser_changelist')
```
2ynn, line 279: Note: using `reverse('admin:member_authuser_change', user.id) to get the full url right away would raise circular import issue

saskatoon/member/admin.py
```diff
@@ -259,13 +270,35 @@ class PersonAdmin(admin.ModelAdmin):
...
 
+    @admin.action(description="Create missing user(s)")
+    def create_user(self, request, queryset):
+        """Associate Person with new AuthUser object"""
+        nb_users = 0
+        for person in queryset.filter(auth_user__isnull=True):
+            email = f"FAKE@person.{person.pk}"
+            AuthUser.objects.create(email=email, person=person)
+            nb_users += 1
+        messages.add_message(request, messages.SUCCESS,
```
tristanlatr, line 297: This looks useful, but I wonder if we're not end up with many FAKE authuser in the DB. It's probably better not have them at all I supose.

tristanlatr (approved):
Thanks for this @2ynn !

saskatoon/member/admin.py
2ynn: The problem when a Person is not associated with a AuthUser is that their email cannot be set, which is what core members seemed to complain about (i.e. owner's email only showing in the comments). We could look into a special button to create this AuthUser on demand (only when a person without an email needs its' email field to be updated) but honestly I think we should be consistent throughout the database and each Person should be associated with an AuthUser. Only pickleader and core members will have password sets though.

2ynn: I should also probably find a way to look for `pending_contact_email` values that would be hidden in some Property models (pending fields only show when property is pending) I'm sure there are some owner's email addresses that were not properly entered because the process to create a Person/AuthUser pair was not clear (looks like the comment section was used for this most of the time).

PR #316, opened 2022-06-18
Fix 315 Make publish_date optional and simplify is_publishable

### LINKED ISSUE #315: Publication date does not get set when saving HarvestForm
## Type of Issue: **Bug report**

On server, `publication_date` is always None no matter what harvest status is

## Steps to Reproduce:
1. Edit a harvest: using HarvestForm, 
3. On save, publication_date (hidden input field) should be set to `now` if status is [ready, scheduled or succeeded] (see `harvest.forms.HarvestForm.clean_publication_date`
4. Go to admin panel to check if the publication_date is still None.

This issue currently occurs on the server but **I can't reproduce it locally.**

As a result:
1. harvests don't show publicly in the calendar (because `is_publishable` is False)
2. Requests for participation are rejected (because `is_open_to_requests` is False)

Note: the publication_date could be a useful feature to control in advance when we would like to publish a harvest (if it were to be made visible in the HarvestForm for core members to edit). As it is now  though it only acts as a proxy for the harvest status. We could consider changing the `is_publishable` harvest property to something more straightforward and not rely on `publish_date`, e.g.:

```
    def is_publishable(self):
        return self.status in ["Ready", "Date-scheduled", "Succeeded"]

    def is_open_to_requests(self):
        return self.status == "Date-scheduled" and timezone.now() < self.start_date
```

BTW `is_open_to_requests` is currently true the whole day when the harvest is happening (comparing dates, not datetimes)

```
def is_open_to_requests(self):
        now = datetime.datetime.now().date()
        start_date = self.start_date.date()
[...]

```
 we should clarify this, for example close requests X hours before harvest start_date.

## Related URLs:
https://saskatoon.lesfruitsdefendus.org/harvest/update/<id>/
https://saskatoon.lesfruitsdefendus.org/admin/harvest/harvest/<id>/change/

2ynn:
Fixes #315

**Urgent, targeting master**
----------
## Type of change:
- [x] Bug fix (change which fixes an issue).
- [ ] Documentation change.
----------
## What's Changed:
- `publication_date` now appears has an optional field in Harvest Form
- if it is left empty (`publication_date=None`), then `is_publishable` simply checks for the harvest status
- if it is set, then we take it into account in `is_publishable` as done before
- Calendar used to compare dates and not datetimes, meaning harvest could still be opened the day of and I believe the ("calendar updates at 18h" (0:00 UTC) limitation was also related to this, which is unecessary.
- Requests are now open until 2 HOURS prior to harvests, we should tweak this number based on core members' needs

--------
## Affected URLs:
http://localhost:8000/harvest/update/<id>

-------
## Checklist:
- [x]] My changes generate no new warnings.

### DIFF
--- a/saskatoon/harvest/forms.py
+++ b/saskatoon/harvest/forms.py
@@ -518,6 +518,7 @@ class HarvestForm(forms.ModelForm):
             'pick_leader',
             'start_date',
             'end_date',
+            'publication_date',
             'nb_required_pickers',
             'about',
         )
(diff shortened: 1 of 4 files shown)


2ynn:
I mean it,s all necessary to fix the harvests not showing in the public calendar. is_urgent and is_happening are not really being used, its mostly just cleaning up.
What do you think about closing requests 2 hours prior to harvest? Should I put way more?

qynn



Jun 18, 2022, 8:19 PM by ***@***.***:

mina-andajos (approved):
### Thanks qynn for the quick fix!
the form and calendar seem to be working well:
image
empty publish_date, status changed to open.
----------------
set the start time to 1 hour from now, can't participate, changed to 3 hours and rfps work:
image
------------------------
All looks good to me.

PR #346, opened 2022-08-10
Implement REST pagination

### LINKED ISSUE #55: Add real tabulation for all list views - Avoid Redis
As the database grows, we're getting some import lag on loading the list views.

We currently use a same DRF-based view for delivering both html and json. This has some advantages, but as we get away from Django native views, we start getting trouble integrating to other libraries. That said, I couldn't make a server-side pagination work with our DRF views and filters together with the DataTables js library. It's certainly possible to do, but I just failed so far :-)

I tried playing with native Django pagination. I also tried a few libraries, such as:

- https://github.com/izimobil/django-rest-framework-datatables
- https://github.com/sheikhfaisalanwar/django-rest-framework-datatables-filter

mina-atef-00:
Fixes #55 
----------
## Type of change:
- [x] New feature.

----------
## What's Changed:
- d5c7509de0b722fba130dfe4b16a1ca0d52890f7 fixed `model.harvest.get_neighborhood()` as it was causing an issue with viewing json output.
- 9a8e813b31a02eac8f5f0ce2f411ef8d5e103104 created a custom pagination class inheriting from `PageNumberPagination`, adding custom `get_paginated_response()` & query params.
- 239c6d494010990ffdc1f37e0b6c108d3e0184b7 changed the context of all list views to implement pagination.
- c90047055fcace19156dab3c46b90f34fb189850 added a custom template filter to modify the request url when changing page size: it removes `page` and `page_size` params from the url (`_clean_url()`), and then adds `page_size=` depending on whether the url has query params after the cleaning or not. Will be used in the entries dropdown in the template.
- 9f25a84f6f87c0e6f0d5a4cd15702098658c3496 modified the base list view template: removed the datatable class, added pagination and custom page size buttons.
- c0af6ae4734b7b91b69cee48ad6351d583ff7172 added back some data-table elements like search and ordering, changed the styling of the buttons I created accordingly.
- 406d132d4abbf4e2d8407117355478129cbe03ac added `renderer_format_needs_json_response()` to all list views to return the json response without a custom context incase the json format is requested or the drf browsable api is used.

--------
## Screenshots:
1. 
image

2. 
image
3. 
image

--------
## Affected URLs:
- http://127.0.0.1:8000/harvest/
- all list views

-------
## Checklist:
- [x] I have made corresponding changes to the documentation.

### DIFF
--- a/saskatoon/saskatoon/settings.py
+++ b/saskatoon/saskatoon/settings.py
@@ -228,7 +228,8 @@ REST_FRAMEWORK = {
    'DEFAULT_RENDERER_CLASSES': (
         get_drf_template_mode(),
         'rest_framework.renderers.JSONRenderer',
-    )
+    ),
+    'DEFAULT_PAGINATION_CLASS': 'saskatoon.pagination.BasicPageNumberPagination',
 }
 
 CACHES = {
--- a/saskatoon/sitebase/templatetags/pagination.py
+++ b/saskatoon/sitebase/templatetags/pagination.py
@@ -1,0 +1,47 @@
+from django import template
+from rest_framework.utils.urls import remove_query_param, replace_query_param
+from saskatoon.pagination import BasicPageNumberPagination
+
+register = template.Library()
+page_query_param = BasicPageNumberPagination.page_query_param
+page_size_query_param = BasicPageNumberPagination.page_size_query_param
+
+
+@register.filter
+def add_page_size_param(url: str) -> str:
+    """adds page_size query parameter to the full request url"""
+    try:
+        url = _clean_url(url)
+        url_has_params: bool = "?" in url
+
+        if url_has_params:
+            return "{}&page_size=".format(url)
+        else:
+            return "{}?page_size=".format(url)
+    except:
+        return url
+
+
+def _clean_url(url: str) -> str:
+    """Removes the page and page_size query parameters"""
+    url = remove_query_param(url, page_query_param)
+    url = remove_query_param(url, page_size_query_param)
+    return url
+
+
+@register.filter
+def get_pages_range(page_number: int) -> list:
+    """Returns a list containing numbers leading to the page_number param"""
+    try:
+        return [i for i in range(1, page_number+1)]
+    except:
+        return [page_number]
+
+
+@register.simple_tag
+def get_page_url(url: str, page_number: int) -> str:
+    """Replaces the page number query parameter of the url"""
+    try:
+        return replace_query_param(url, page_query_param, page_number)
+    except:
+        return '#'
--- a/saskatoon/harvest/api.py
+++ b/saskatoon/harvest/api.py
@@ -60,12 +60,22 @@ class HarvestViewset(LoginRequiredMixin, viewsets.ModelViewSet):
         if request.accepted_renderer.format == 'json':
             return Response(response.data)
         # default request format is html:
-        return Response({'data': response.data,
-                         'filter': get_filter_context(self),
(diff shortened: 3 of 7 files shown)


mina-andajos:
> Thank you @mina-andajos for this heavy work.
> 
> > and the div is generated by the datatable code, so I don't know how I can change the element so I can fit the entries button in.
> 
> If the dug is generated by data table, you would need to use JS script to I just your html. But maybe that would be cumbersome. Another option we have is to tweak the css with negative margin or something to show both div on the same level. Maybe that the easier.
> 
> 
> Tell me what you think.
> 

@tristanlatr Yeah I was thinking I had to make a js script to make this work. I will try it out and see. thanks for the help!

saskatoon/saskatoon/settings.py
```diff
@@ -225,7 +225,8 @@ def get_drf_template_mode() -> str:
    'DEFAULT_RENDERER_CLASSES': (
         get_drf_template_mode(),
```
2ynn, line 226: @mina-andajos if I set "DRF_BROWSABLE_API_MODE=yes" in .env I get `Object of type HarvestFilterForm is not JSON serializable` when trying to GET the harvest page. Is this expected behavior? If not should we drop drf browser mode?

saskatoon/sitebase/templatetags/pagination.py
```diff
@@ -0,0 +1,47 @@
+from django import template
+from rest_framework.utils.urls import remove_query_param, replace_query_param
+from saskatoon.pagination import BasicPageNumberPagination
+
+register = template.Library()
+page_query_param = BasicPageNumberPagination.page_query_param
+page_size_query_param = BasicPageNumberPagination.page_size_query_param
```
2ynn, line 7: how about importing these two values like this ?
```
from saskatoon.pagination.BasicPageNumberPagination import page_query_param, page_size_query_param
```

2ynn (commented):
nice work!

saskatoon/saskatoon/settings.py
mina-andajos: it's not no
I think it just needs a quick fix with the serializer. I will check it out.

mina-andajos: the forms that we pass to the context aren't serializeable. I remember discussing this before. I could add an extra condition that if the template renderer is the drf one then we could pass just the json response without a custom context

mina-andajos: something like this:
```py
    def list(self, request, *args, **kwargs):
        self.template_name = 'app/list_views/harvest/view.html'
        self.serializer_class = HarvestListSerializer

        response = super(HarvestViewset, self).list(request, *args, **kwargs)
        if request.accepted_renderer.format in ('json', 'api'):
            return Response(response.data)
```

saskatoon/sitebase/templatetags/pagination.py
mina-andajos: it refuses to import it like that,
it says `'saskatoon.pagination' is not a package` althought the saskatoon folder has `__init__.py`

full exception:
```sh
django.template.library.InvalidTemplateLibrary: Invalid template library specified. ImportError raised when trying to load 'sitebase.templatetags.pagination': No module named 'saskatoon.pagination.BasicPageNumberPagination'; 'saskatoon.pagination' is not a package
```

2ynn: ok my bad!!

2ynn (approved):

PR #367, opened 2023-09-29
fix integrity error on get_or_create authuser

2ynn:
When saving a `PersonUpdateForm`, if email and roles are provided but the person is not yet associated with an AuthUser, one should be created automatically. The `get_or_create` method called during `save` will first assign an empty string for the email, which then  gets updated right after (passing email to `get_or_create` would break the flow of changing an existing auth_user email). However if the database happens to have an AuthUser with an empty string as email an Integrity error is thrown when `get_or_create` tries to create a new one.

This PR avoids this method and makes use of the fact `self.auth_user` was already stored during `__init__`

## Type of change:
- [x] Bug fix (change which fixes an issue).
- [ ] Documentation change.
----------
## What's Changed:
- fix create auth_user if it does not exist in `PersonUpdateForm`
- make sure email address is provided if a role is selected
- make sure empty email does not trigger duplicate error
- cosmetic changes on form validation strings

--------
## Affected URLs:
e.g. `localhost:8000/person/update/pk/?pid=pid`

-------
## Checklist:

### DIFF
--- a/saskatoon/harvest/forms.py
+++ b/saskatoon/harvest/forms.py
@@ -300,7 +300,7 @@ class PropertyCreateForm(PropertyForm):
                 validate_email(data['owner_email'])
             else:
                 raise forms.ValidationError(
-                    _("ERROR: You must either select an Owner \
+                    _("You must either select an Owner \
                     or create a new one and provide their personal information"))
         return data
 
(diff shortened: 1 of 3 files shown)


2ynn:
thanks @tristanlatr
Sorry I didn't realize it was still Draft. I'd be ready to merge

PR #383, opened 2024-04-08
fix #117 make homepage wysiwygly editable from admin panel

### LINKED ISSUE #117: Add editable HomePage model

2ynn:
Addresses #117 

- added `Content` model with title/subtitle/body fields (en/fr)
- installed `django-quill` for WYSIWYG editing

screenshot_2024-04-08_00:54:07
screenshot_2024-04-08_00:53:56


saskatoon/sitebase/views.py
```diff
@@ -7,11 +7,29 @@
...
+
+        content_name = VOLUNTEER_HOME_CONTENT_NAME
+
+        if self.request.user.is_authenticated:
+            content_name = PICKLEADER_HOME_CONTENT_NAME
+
+        home = Content.objects.get(name=content_name)
+        if home is not None:
+            context['content'] = home.content(self.request.LANGUAGE_CODE)
+
```
2ynn, line 30: TODO: pass a dummy content object here if there are no appropriate object saved in db

saskatoon/sitebase/templates/app/index.html
```diff
@@ -5,7 +5,7 @@
 {% block content %}
 
 <div class="breadcomb-area">
-    <div class="container mg-t-30">
+    <div class="container">
```
2ynn, line 8: TODO: check if `mg-t-30` class is needed

saskatoon/sitebase/views.py
oxgr: my lsp is saying `Content.objects` doesn't exist. as it stands, hitting `/` route gives this error. wrapping the method in try/catch and manually setting `home = None` on exception at least bypasses this, but original problem remains and `get()` may never get called. not sure where to go from there.

2ynn: have you run migrations? `saskatoon/manage.py migrate`

2ynn: pushed new commit, now if the `Content` object with reference name `pickleader_home`/`volunteer_home` doesn't exist when first accessing the home page, a blank object is created.

oxgr:
loads good with fresh db and fixtures! looks good to go?

oxgr (approved):

PR #385, opened 2024-04-13
Remove ckeditor dependency

2ynn:
PR #383 introduced `django-quill` as WYSIWYG editor even though `ckeditor` was already used elsewhere (`harvest.about` form widget). Since `ckeditor` version 4 is deprecated, it is replaced by django-quill to avoid multiple wysiwyg editor libraries being required.

WARNING: to migrate the `harvest.about` field to `QuillField`, this extra step must be taken:
```
python saskatoon/manage.py convert_to_quill harvest Harvest about
```
----------
## Type of change:
- [ ] Documentation change.
----------


oxgr (approved):
had to do a data dump that to convert harvest fixtures to json that quill can parse so tests pass. fixtures still loaded for me before running `init` though, so not sure if the parsing error only shows up on `loaddata`.

either way, looks good now, but might have to keep in mind when adding new content fixtures

2ynn:
> had to do a data dump that to convert harvest fixtures to json that quill can parse so tests pass. fixtures still loaded for me before running `init` though, so not sure if the parsing error only shows up on `loaddata`.

I'm not sure I understand, are the fixtures provided in this PR working with `loaddata` ?

2ynn:
> I'm not sure I understand, are the fixtures provided in this PR working with `loaddata` ?
Ok my bad, did not see the last commit. Thanks for updating the fixture, good catch!

Issue #372, opened 2024-03-12
Improve pickleader onboarding flow


2ynn:
Pickleaders should read and agree to a confidentiality agreement before  being granted access to the system

2ynn:
related to PR #230

2ynn:
A new `Onboarding` model will be created (OneToMany relationship with Person model).
From the admin panel, it should be easy to enter multiple email address at once.
Once an *onboarding group* is ready, an *invitation* can be sent using an admin action: this will automatically send an email to all newly registered users and  provide them with a temporary password. 

When they first log into the platform, onboarding pickleaders are marked as *pending validation* by being assigned the `volunteer` role. They will first be redirected to a terms&condition page that they need to sign. Agreeing to the form will update their role to `pickleader`, which will give them access to a Person Update form so they can update their personal information

TODO:
- [ ] set up Onboarding model + admin forms 
- [ ] admin action: generate temporary password + send email invite
- [ ] set up terms& condition page using generic `Content` model
- [ ] set up redirect to Person Update form

oxgr:
A new `password_set` field will be added to `AuthUser` to differentiate newly-onboarded users and existing pickleaders looking change password.

`Index` will redirect based on `volunteer` role with password and `AuthUser.password_set` value.

Some extra steps added to the pickleader onboarding flow:

1. First-time login

- with a temp password provided by email
- members only assigned a `volunteer` role.

2. Terms & conditions

- agrees => proceeds to next step
- disagrees => log out

3. Person update form

- hide roles, email
- keep `Mixin`, override `has_permission()` method in view for perms
- on save, add a `pick_leader` role and keep `volunteeer`

4. Reset password

- a few rules of password complexity
- on save, set `password_set` field to `True`

Existing pickleaders should be able to reset their password by asking an admin (revolving icon beside emails in Community view). This will set `password_set` to `False` which, on login, automatically redirects members to `reset_password` url.

TODO:

- [x] set up Onboarding model + admin forms - #230
- [ ] admin action: generate temporary password + send email invite
- [ ] set up terms& condition page using generic Content model
- [ ] set up redirect to Person Update form + password reset

PR #386, opened 2024-04-14
feat: implement Onboarding model + admin inline forms

2ynn:
Partially addresses #372 

- [X] set up Onboarding model + admin forms 
- [ ] admin action: generate temporary password + send email invite
- [ ] set up terms& condition page using generic `Content` model
- [ ] set up redirect to Person Update form

screenshot_2024-04-14_19:29:34


oxgr (approved):
cool! was considering an inverse filter to have a list for all onboarded pickleaders that never set password as candidates for some sort of reminder email, but that can be solved by using "active" filter.

2ynn:
> cool! was considering an inverse filter to have a list for all onboarded pickleaders that never set password as candidates for some sort of reminder email, but that can be solved by using "active" filter.

@oxgr no I don't think we can use the `active` boolean. I believe if a user is set as `active=False` they wouldn't be able to log in.
Again the way we check if a new pickleader has not finished onboarding is by checking that they still have the 'volunteer' role. We could make it so their role is only set to `pickleader` once they have updated their password

PR #391, opened 2024-04-26
feat: redirect to password change view

oxgr:
On `Index` view, redirect to a password change view if `has_temporary_password` field is set to `True`.

Sets to `False` on successful submit.

Also sets `has_temporary_password` to `True` for new onboarded users when sending first login email.

Partially addresses #372 

## Type of change:

- [ ] Documentation change.

---

## What's Changed:

- add `password_set` field to `AuthUser`
- add Password Change view
- redirect to Password Change if `password_set == False`
- set `password_set` for new users in first login email
- remove old "Change Password" tab in user dropdown (inaccessible without admin permissions)

---

## Screenshots:
20240426_04h22m49s_grim

---

## Affected URLs:

- `127.0.0.1:8000/`
- `127.0.0.1:8000/change_password/`

---

## Checklist:


saskatoon/member/admin.py
```diff
@@ -96,6 +96,7 @@ class AuthUserAdmin(UserAdmin):
                     'is_core',
                     'is_admin',
                     'is_active',
+                    'password_set',
                     'id',
                     'date_joined',
                     'has_password',
```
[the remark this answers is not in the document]
2ynn: alternatively a custom filter could be good as well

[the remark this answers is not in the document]
oxgr: a filter was already set up! just had to add a check for `password_set`

i tried adding an "Onboarded pickleader" lookup, but it was pretty much the same as a regular pickleader filter

saskatoon/member/forms.py
```diff
@@ -209,3 +212,21 @@ def save(self):
...
+
+class PasswordChangeForm(auth_forms.PasswordChangeForm):
+
+    def __init__(self, user, *args, **kwargs):
+        super().__init__(user, *args, **kwargs)
+
+        # replace widgets with placeholder values to fit Notika theme
+        self.fields['old_password'].widget = PasswordInput(attrs={'class': 'form-control', 'autocomplete': 'current-password', 'placeholder': 'Old password', 'autofocus': True })
+        self.fields['new_password1'].widget = PasswordInput(attrs={'class': 'form-control', 'autocomplete': 'new-password', 'placeholder': 'New password'})
+        self.fields['new_password2'].widget = PasswordInput(attrs={'class': 'form-control', 'autocomplete': 'new-password', 'placeholder': 'Confirm password'})
```
2ynn, line 225: can you try and set the `widgets` field instead? see [OrganizationForm](https://github.com/LesFruitsDefendus/saskatoon-ng/blob/9eee8ffba13f0978916fab800ec3970be38a8737/saskatoon/member/forms.py#L138) for example

saskatoon/member/forms.py
```diff
@@ -209,3 +212,21 @@ def save(self):
...
+
+    def save(self, commit=True):
+        instance = super().save(commit)
+        auth_user = AuthUser.objects.get(person=instance.person)
+        auth_user.password_set = True
+        auth_user.save()
```
2ynn, line 231: I think this should work: 
```suggestion
        person.auth_user.password_set = True
        person.auth_user.save()
```

saskatoon/member/views.py
```diff
@@ -113,3 +117,16 @@ def get_context_data(self, **kwargs):
...
         return context
+
+
+class PasswordChangeView(auth_views.PasswordChangeView):
+    template_name = 'registration/change_password.html'
+    form_class = PasswordChangeForm
+
+    def get_context_data(self, **kwargs):
+        context = super(PasswordChangeView, self).get_context_data(**kwargs)
+        return context
```
2ynn, line 128: What is this doing? looks like you don't need it

saskatoon/sitebase/templates/registration/base.html
```diff
@@ -0,0 +1,83 @@
+{% load i18n %}
```
2ynn, line 1: Just curious, do we really need to create a new base? why not extending from app/base/view.html?
Are stylesheets very different from the ones needed in the app?

saskatoon/sitebase/templates/registration/change_password.html
```diff
@@ -0,0 +1,65 @@
+{% extends 'registration/base.html' %}
```
2ynn, line 1: what prevented you from being able to reuse a basic form base like https://github.com/LesFruitsDefendus/saskatoon-ng/blob/develop/saskatoon/sitebase/templates/app/forms/base.html ? Can you add a couple screenshots to explain what  your workaround is achieving for displaying form errors?

saskatoon/sitebase/templates/registration/login.html
```diff
@@ -104,10 +35,10 @@
...
                 </div>
 
-                <div class="nk-navigation nk-lg-ic">
-                    <!-- <a href="#" data-ma-action="nk-login-switch" data-ma-block="#l-register"><i class="notika-icon notika-plus-symbol"></i> <span>{% trans 'Register' %}</span></a> -->
-                    <a href="{% url 'reset_password' %}" data-ma-action="nk-login-switch" data-ma-block="#l-forget-password"><i>?</i> <span>{% trans "Forgot Password" %}</span></a>
-                </div>
+                <!-- <div class="nk-navigation nk-lg-ic"> -->
+                <!--     <a href="#" data-ma-action="nk-login-switch" data-ma-block="#l-register"><i class="notika-icon notika-plus-symbol"></i> <span>{% trans 'Register' %}</span></a> -->
+                <!--     <a href="{% url 'reset_password' %}" data-ma-action="nk-login-switch" data-ma-block="#l-forget-password"><i>?</i> <span>{% trans "Forgot Password" %}</span></a> -->
+                <!-- </div> -->
```
2ynn, line 41: please delete if no longer needed

saskatoon/sitebase/views.py
```diff
@@ -30,6 +31,13 @@ def get_context_data(self, **kwargs):
 
         return context
 
+    def dispatch(self, request, *args, **kwargs):
+
+        if self.request.user.is_authenticated and self.request.user.password_set == False:
```
2ynn, line 36:
```suggestion
        if self.request.user.is_authenticated and not self.request.user.password_set:
```

2ynn (changes requested):

saskatoon/member/forms.py
oxgr: `widgets` field is inherited from `ModelForm`. this class extends `django.contrib.auth.PasswordForm` which very lightly extends `BaseForm`.

If for legibility, I could try to simplify with a `map`?

saskatoon/member/forms.py
oxgr: nice. even simpler - i used `instance.password_set`

saskatoon/member/views.py
oxgr: true! removing...

saskatoon/sitebase/templates/registration/base.html
oxgr: i just moved existing tags in `login.html`, didn't realise most of the tags were the same as `base/view.html`.

refactored `base` to make a template with an empty body. let me know if that works

saskatoon/member/forms.py
2ynn: ok all good

saskatoon/member/forms.py
2ynn: ok I didn't realize `instance` was of type `AuthUser`,  in that case I would prefer:
```
 def save(self, commit=True):
        self.instance.password_set = True
        super().save(commit)
```

saskatoon/sitebase/templates/registration/change_password.html
oxgr: I chose to use the existing `PasswordChangeForm` because it was already there and did most of the heavy lifting.

Workaround was to avoid rendering the fields rendered as text in `ul` elements. Because the text isn't inside any other child element, i couldn't style it with `display: none;`. Wasn't sure how else to go about hiding the arbitrary field names either through templates or form.

here's what it would look like with a simple `{{ form.errors }}`

20240430_16h10m39s_grim
20240430_16h12m03s_grim

saskatoon/member/forms.py
oxgr: `self` is of type `PasswordChangeForm` though. i think AuthUser `instance` is returned from `save()`, hence the initial `save()` call.

saskatoon/member/forms.py
```diff
@@ -209,3 +212,20 @@ def save(self):
...
+        # replace widgets with placeholder values to fit Notika theme
+        self.fields['old_password'].widget = PasswordInput(attrs={'class': 'form-control', 'autocomplete': 'current-password', 'placeholder': 'Old password', 'autofocus': True })
+        self.fields['new_password1'].widget = PasswordInput(attrs={'class': 'form-control', 'autocomplete': 'new-password', 'placeholder': 'New password'})
+        self.fields['new_password2'].widget = PasswordInput(attrs={'class': 'form-control', 'autocomplete': 'new-password', 'placeholder': 'Confirm password'})
+
+    def save(self, commit=True):
+        self.instance.password_set = True
+        return super().save(commit)
+        instance.save()
+        return
```
2ynn, line 231:
```suggestion
    def save(self, commit=True):
        instance = super().save(False)
        instance.password_set = True
        super().save(commit)
```

saskatoon/member/forms.py
2ynn: Oh okay true `self.instance` might be limited to `ModelForm` which `PasswordChangeForm` doesn't inherit from. fixed my suggestion

saskatoon/sitebase/templates/registration/change_password.html
```diff
@@ -0,0 +1,65 @@
...
+      }
+    </style>
+    <div class="login-content">
+        <!-- Login -->
+        <div class="nk-block toggled" id="l-login">
+            <form action='{% url "change_password" %}' method="post">
+                {% csrf_token %}
+                <div class="nk-form">
+                    <h1>Change password</h1>
+                    {% block form %}
```
2ynn, line 17: Are the `{% block form %}` tags doing anything here?  
in `apps/forms/base.html` for example, they are used as placeholders so other templates extending it can define the block. I don't think this is what you are doing here
https://github.com/LesFruitsDefendus/saskatoon-ng/blob/9eee8ffba13f0978916fab800ec3970be38a8737/saskatoon/sitebase/templates/app/forms/base.html#L33

saskatoon/sitebase/templates/registration/change_password.html
```diff
@@ -0,0 +1,65 @@
...
+                                            gap: 1em">
+                                    <!-- Django renders errors with the attr keys as text inside ul component. -->
+                                    <!-- This is a workaround to avoid writing a new template for error rendering. -->
+                                    {% if form.errors.old_password %}<div>{{ form.errors.old_password }}</div>{% endif %}
+                                    {% if form.errors.new_password1 %}<div>{{ form.errors.new_password1 }}</div>{% endif %}
+                                    {% if form.errors.new_password2 %}<div>{{ form.errors.new_password2 }}</div>{% endif %}
+                                </div>
+                                <br>
+                            </div>
+                        {% endif %}
```
2ynn, line 34: have you tried using simply `{{form |crispy}}` ?
see example: https://github.com/LesFruitsDefendus/saskatoon-ng/blob/9eee8ffba13f0978916fab800ec3970be38a8737/saskatoon/sitebase/templates/app/forms/model_form.html#L6

2ynn (changes requested):

saskatoon/sitebase/templates/registration/base.html
2ynn: ok sweet, how about we call your barebone base `app/base/base.html` instead of `view_empty` ?

saskatoon/sitebase/templates/registration/change_password.html
2ynn: I think ` {{ form|crispy }} ` should do that for you, see other comment

saskatoon/sitebase/templates/registration/change_password.html
oxgr: definitely not. i think i just copied existing login form. ill clean that one up too.

saskatoon/sitebase/templates/registration/change_password.html
oxgr: not sure how `crispy works, but doesn't seem to work with errors.

errors render as unordered lists `<ul>...<li>`

saskatoon/sitebase/templates/registration/base.html
oxgr: done!

saskatoon/sitebase/templates/registration/change_password.html
oxgr: switched to crispy and added basic styling to match
20240503_17h42m21s_grim

saskatoon/member/filters.py
```diff
@@ -41,7 +41,7 @@ def queryset(self, request, queryset):
         if self.value():
             # WARNING: Conditions must match `AuthUser.is_onboarding`. Using QuerySet over lists for performance.
             group = Group.objects.get(name='volunteer')
-            return queryset.filter(groups__in=[group]).exclude(password__exact='')
+            return queryset.filter(groups__in=[group]).exclude(password__exact='').filter(password_set=False)
```
2ynn, line 44: Should we also add this check to `is_onboarding` property in AuthUser model  ?

saskatoon/sitebase/templates/app/base/header.html
```diff
@@ -1,99 +1,125 @@
...
+                             alt="Les Fruits Defendus - Saskatoon" />
+                    </a>
                 </div>
             </div>
             <div class="col-md-5 col-xs-12">
                 <div class="logo-area">
                     <h2 style="color:white">Les Fruits Défendus</h2>
-                    <p style="color:white"><i>{% translate 'Collectif montréalais de cueillette urbaine de fruits' %}</i></p>
+                    <p style="color:white">
+                        <i>{% translate 'Collectif montréalais de cueillette urbaine de fruits' %}</i>
```
2ynn, line 19: If we want this bilingual we should set it in english and mark it for translation.
Alternatively we could remove the translate tags and keep it in french as it's the logo

saskatoon/sitebase/templates/app/base/header.html
```diff
@@ -1,99 +1,125 @@
...
+                                        </li>
+                                    {% endif %}
+                                    {% if perms.member.change_organization %}
+                                        <li>
+                                            <a href="{% url 'admin:member_organization_changelist' %}">{% translate "Manage beneficiary" %}</a>
+                                        </li>
+                                    {% endif %}
+                                    <li>
+                                        <a href="{% url 'admin:index' %}">{% translate "Admin panel" %}</a>
+                                    </li>
```
2ynn, line 74: i'd say let's move this inside the {% if user|is_core_or_admin %} check

saskatoon/sitebase/templates/app/base/header.html
```diff
@@ -1,99 +1,125 @@
...
+                                </li>
+                            {% endfor %}
+                        </ul>
+                    </li>
+                </ul>
             </div>
         </div>
     </div>
 </div>
+</div>
```
2ynn, line 125: is this extra </div> tag intentional?

saskatoon/sitebase/views.py
```diff
@@ -36,8 +38,14 @@ def dispatch(self, request, *args, **kwargs):
         """Redirect new pickleaders"""
 
         user = self.request.user
-        if user.is_authenticated and user.is_onboarding:
+
+        if user.is_authenticated:
+            if not user.password_set:
+                return redirect('change_password')
```
2ynn, line 44: shouldn't this check come after the `is_onboarding` ? iiiuc an onboarding pickleader has `password_set = False`

2ynn (changes requested):

saskatoon/member/admin.py
```diff
@@ -96,9 +96,9 @@ class AuthUserAdmin(UserAdmin):
                     'is_core',
                     'is_admin',
                     'is_active',
+                    # 'has_temporary_password',
                     'id',
                     'date_joined',
-                    'has_password',
```
2ynn, line 101: `has_password` still have value in list_display i think.

saskatoon/member/admin.py
```diff
@@ -96,9 +96,9 @@ class AuthUserAdmin(UserAdmin):
                     'is_core',
                     'is_admin',
                     'is_active',
+                    # 'has_temporary_password',
```
2ynn, line 99: please remove comment, I think the `UserIsOnboarding` filter is enough, no need for `has_temporary_password` in list_display

saskatoon/member/models.py
```diff
@@ -73,6 +73,9 @@ class AuthUser(AbstractBaseUser, PermissionsMixin):
 
     is_staff = models.BooleanField(default=False, null=False)
 
+    # Default set to False for backwards compatibility
```
2ynn, line 76: Please remove comment, I find it confusing: default is set to False because new users don't have a temporary password (e.g. a volunteer or owner being added) not only backwards compatibility

saskatoon/member/models.py
```diff
@@ -109,8 +112,9 @@ def is_onboarding(self):
         # Retrieve only name fields from QuerySet of Groups
         group_names = [g.name  for g in self.role_groups]
         return ('pickleader' not in group_names and
-            'volunteer' in group_names and
-            self.password != '')
+                'volunteer' in group_names and
+                self.has_temporary_password and
+                self.password != '')
```
2ynn, line 117: if `self.has_temporary_password` is `True`, wouldn't that imply that `self.password != ''` ? 

```
return (
    'pickleader' not in group_names and
    'volunteer' in group_names and
    self.has_temporary_password
)
```

saskatoon/sitebase/views.py
```diff
@@ -36,8 +38,14 @@ def dispatch(self, request, *args, **kwargs):
...
         user = self.request.user
-        if user.is_authenticated and user.is_onboarding:
+
+        if user.is_authenticated:
+            if user.is_onboarding:
                 return redirect('terms_conditions')
+
+            if user.has_temporary_password:
+                return redirect('change_password')
+
```
2ynn, line 48:
```suggestion

        if not user.is_authenticated:
            return super().dispatch(request, *args, **kwargs)
            
        if user.is_onboarding:
            return redirect('terms_conditions')

        if user.has_temporary_password:
            return redirect('change_password')
```

saskatoon/member/forms.py
```diff
@@ -220,3 +223,20 @@ def save(self):
...
+    def __init__(self, user, *args, **kwargs):
+        super().__init__(user, *args, **kwargs)
+
+        # replace widgets with placeholder values to fit Notika theme
+        self.fields['old_password'].widget = PasswordInput(attrs={'class': 'form-control', 'autocomplete': 'current-password', 'placeholder': 'Old password', 'autofocus': True })
+        self.fields['new_password1'].widget = PasswordInput(attrs={'class': 'form-control', 'autocomplete': 'new-password', 'placeholder': 'New password'})
+        self.fields['new_password2'].widget = PasswordInput(attrs={'class': 'form-control', 'autocomplete': 'new-password', 'placeholder': 'Confirm password'})
+
+    def save(self, commit=True):
+        instance = super().save(False)
```
2ynn, line 239: just curious, what happens if `old_password == new_password1 == new_password2` ? does the form throw an error?

saskatoon/member/admin.py
```diff
@@ -147,6 +147,7 @@ def has_password(self, user):
                     'is_active',
                     'is_staff',
                     'is_superuser',
+                    # 'has_temporary_password',
```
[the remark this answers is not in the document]
oxgr: not needed - removed

saskatoon/member/admin.py
oxgr: removed

saskatoon/member/models.py
oxgr: true, i just kept stacking conditions.

saskatoon/member/forms.py
oxgr: no error - good catch! i just left the password checks to `auth.PasswordChangeForm` and `AUTH_PASSWORD_VALIDATORS` in `settings.py`. 

I'll see where I can fit another vaildator

saskatoon/sitebase/templates/app/base/header.html
oxgr: I think logo can be kept in French. It's a nice little touch

saskatoon/sitebase/templates/app/base/header.html
oxgr: good catch. the extra tag was actually an `<a>` and my formatter skipped the indent for it

saskatoon/member/forms.py
oxgr: added password unchanged validator

saskatoon/member/forms.py
```diff
@@ -220,3 +223,38 @@ def save(self):
...
+        self.fields['new_password2'].widget = PasswordInput(attrs={'class': 'form-control', 'autocomplete': 'new-password', 'placeholder': 'Confirm password'})
+
+    def clean_new_password1(self):
+        """
+        Validate that the new_password does not match old_password.
+        """
+        old_password = self.cleaned_data.get('old_password')
+        new_password = self.cleaned_data.get('new_password1')
+        if old_password and new_password:
+            if old_password == new_password:
```
2ynn, line 249: How about simply:
```suggestion
        if new_password is not None and old_password == new_password:
```

2ynn (approved):
some minor comments only

PR #396, opened 2024-05-24
Set default value for authorized filter in property-list view

2ynn:
PR #320  introduced workarounds to reduce initial loading time when clicking on menu items despite pagination not being properly implemented. However loading time for the `Properties` is still problematic.

This PR: 
- sets the `Authorized` filter to `Not Yet Authorized` so that less properties are loaded at once when the `Properties` page is accessed from the menu. 
- prevents most redirects to `harvest-list`, `property-list` and `community-list` routes, forcing users to go through the menu buttons.


oxgr (approved):
simple and snappy!

2ynn:
@oxgr I think I am gonna remove redirects to `community-list` as well, iirc timeouts were also experienced when returning from PersonUpdateForm

oxgr:
hiding onboarding field was a good call!

PR #397, opened 2024-05-24
feat: onboarding - agreed_terms field

### LINKED ISSUE #372: Improve pickleader onboarding flow

oxgr:
Adds an explicit `agreed_terms` field to `AuthUser` denoting whether they have agreed to the T&C. Useful for admin tasks when confidentiality agreement is ever updated.

This PR also sends onboarding users to `terms_conditions` and `onboarding-person-update` views on separate redirects from `Index` as the former should be reachable to existing onboarded users.

Ready as-is, but doesn't feel very clean with model logic in a `Form`less view. Right now, `agreed_terms` is set for a user by `POST` to `/terms_conditions/`. Taking suggestions!

Partially fixes #372 

---

## Type of change:

- [ ] Documentation change.

---

## What's Changed:

- added `AuthUser.agreed_terms` field
- agree button in `TermsConditionsView` `POST`s to same URL, altering said field

---

## Affected URLs:

`/`
`/terms_and_conditions/`

---

## Checklist:


saskatoon/sitebase/views.py
```diff
@@ -63,6 +64,12 @@ def get_context_data(self, **kwargs):
 
         return context
 
+    def post(self, *args, **kwargs):
+        self.request.user.agreed_terms = True
+        self.request.user.save()
+        return redirect('home')
```
oxgr, line 70: this feels like a shoddy workaround, but going with a form for a single field felt like overkill.

not sure how to go about this 😖

saskatoon/member/admin.py
```diff
@@ -279,6 +282,13 @@ def clean_column(col: Optional[str]) -> str:
                     f"Something went wrong: {e}"
                 )
 
+    @admin.action(description="Sets selected User(s)'s agreed_terms field to False")
+    def unset_agreed_terms(self, request, queryset):
+        for u in queryset:
+            u.agreed_terms = False
+            u.save()
```
2ynn, line 289: i which context would we need that? in case we update the terms?

saskatoon/member/admin.py
```diff
@@ -164,6 +166,7 @@ def has_password(self, user):
                     'is_staff',
                     'is_superuser',
                     'has_temporary_password',
+                    'agreed_terms',
```
2ynn, line 169: I think you want this field in `fieldsets`, not `add_fieldsets`, essentially if for some reason we create a new user from admin panel then `agreed_terms` get set to False, and then later on we want to be able as admin to toggle the value if needed

saskatoon/sitebase/views.py
2ynn: that's totally fine!!

saskatoon/member/admin.py
2ynn: Btw it's better to avoid the for loop in such a case: `queryset.update(agreed_terms=False)`

2ynn (changes requested):
looks good, just want to make sure admins can manually toggle the `agreed_terms` value

saskatoon/member/admin.py
oxgr: simone mentioned wanting to reset this field for everyone in case the agreement ever updates.

maybe i should rename to `reset_agreed_terms`

I just copied from other actions above, but def gonna go for queryset

2ynn (approved):
looks good!

PR #416, opened 2024-10-23
feat: organization details/lists pages

### LINKED ISSUE #415: Organization (beneficiary and/or equipment point) list and detail pages
## Type of Issue: **Feature request**

Beneficiaries should have their own details page to see more information on a particular organisation.

This will especially be beneficial in the future to view more details on equipment available from an organisation.

## Expected Behavior:

- Beneficiaries listed in the Beneficiary tab should have clickable links to a details page, like in Harvest and Community tabs.

## Current Behavior:

- only a list
- long & unreadable description column (could be removed with a dedicated details page)

## Related URLs:

e.g. `https://saskatoon.lesfruitsdefendus.org/beneficiary/`

oxgr:
Add a details page for an `Organization`, following the same pattern as `Property`.
Also converts the `Equipment` tab into an `Equipment Point` list page, showing `Organization`s with `is_equipment_point`

The details page is accessible via the corresponding `Beneficiary` and `Equipment Point` list page by clicking on the name of the organization.

Fixes #415

---

## Type of change:

---

## What's Changed:

- Organization Details page
  - status for equipment point + beneficiary as icons in header
  - if equipment point, includes a table of equipment at the bottom
- Beneficiary List page
  - Org name links to detail view
  - removed description column
- Organization Create/Update View
  - renamed route from `beneficiary-[create|update]` to `organization-[create|update]`

---

## Screenshots:

20241101_00h55m59s_grim
20241127_17h47m05s_grim

---

## Affected URLs:

`127.0.0.1:8000/beneficiaries/`
`127.0.0.1:8000/equipment-points/`
`127.0.0.1:8000/organization/<int:pk>`

---

## Checklist:


2ynn:
@oxgr for this PR specifically if you can provide some screenshots when it's ready i think it'll ease the review process. Thanks!

saskatoon/member/urls.py
```diff
@@ -8,9 +8,9 @@
          views.PersonCreateView.as_view(),
          name='person-create'),
 
-    path('beneficiary/create/',
+    path('organization/create/',
```
2ynn, line 11: Before changing the actual routes we need to make sure docs don't mention them anywhere, in which case think it could be okay just changing the reference name

saskatoon/sitebase/templates/app/detail_views/organization/equipment_table.html
```diff
@@ -0,0 +1,43 @@
...
+        <div class="data-table-list">
+          <div class="table-responsive">
+            <div>
+              <h4> {% trans Equipment %} </h4>
+              {% if perms.harvest.add_equipment and organization.is_equipment_point %}
+              <a href="{% url 'equipment-create' %}?pid={{ property.id }}">
+                <button title="New Equipment" class="btn btn-success notika-btn-success waves-effect">
+                  <i class="fa fa-plus"></i>
+                  {% trans "New Equipment" %}
+                </button>
```
2ynn, line 16: If you are to add a `New Equipment` button then I would ideally like to prefill and hide the `owner` field from the `EqupmentForm` so we make sure there is no confusion and the newly created equipment actually belongs to the organization. This can be dealt with in a separate PR if needed.

saskatoon/sitebase/templates/app/detail_views/organization/equipment_table.html
```diff
@@ -0,0 +1,43 @@
...
+{% load static %}
+<div class="search-engine-area mg-t-30">
+  <div class="container">
+    <div class="row">
+      <div class="col-lg-12 col-md-12 col-sm-12 col-xs-12">
+        <div class="data-table-list">
+          <div class="table-responsive">
+            <div>
+              <h4> {% trans Equipment %} </h4>
+              {% if perms.harvest.add_equipment and organization.is_equipment_point %}
```
2ynn, line 11: I think if this table is shown `organization.is_equipment_point` is already known to be `True`

2ynn (commented):
thanks! just a few comments

saskatoon/sitebase/templates/app/detail_views/organization/equipment_table.html
2ynn: For reference you can look at how the Property ID `pid` is passed to the `HarvestCreateView` when the `HarvestForm` is opened from within a property detail page:
https://github.com/LesFruitsDefendus/saskatoon-ng/blob/3308f376fa89b210b1be4ba02264f411967d0568/saskatoon/harvest/views.py#L115

2ynn: Another important point is that the `Property` field should be hidden from the form when it is called from an Organization details page.
Altogether I am wondering if a simpler inline form similar to the fruit distribution one would not be preferable here since there really aren't too many fields to fill:
- 1 equipment type autocomplete
- 1 description
- 1 count
https://github.com/LesFruitsDefendus/saskatoon-ng/blob/3308f376fa89b210b1be4ba02264f411967d0568/saskatoon/harvest/views.py#L275

oxgr: added the autofill and took out the `Property` field, but left the `Owner` field for UX. i think it's helpful for user to see what is being implied

i think the separate form is still a good idea for UX consistency. fruit inline form needed a refresh on submit too, so almost defeats the flow.

2ynn: Ok, I that case I suggest we do the following in the `EquipmentForm` : 
- change the label for "Owner" to "Equipment Point"
- change the `actor-autocomplete` for a `equipmentpoint-autocomplete`

saskatoon/harvest/views.py
```diff
@@ -26,6 +27,22 @@ class EquipmentCreateView(PermissionRequiredMixin, SuccessMessageMixin, CreateVi
     success_url = reverse_lazy('equipment-list')
     success_message = _("Equipment created successfully!")
 
+    def get_organization(self):
+        """If creating new harvest from Organization page"""
+        try:
+            pid = self.request.GET['pid']
```
2ynn, line 33: "pid" stood for "Property ID", let's use `aid` ?

oxgr: for `actor_id`? sound good!

oxgr:
@2ynn was going to do this in the next equipment point list page PR, but ended up moving the common `Organization` detail page into a separate viewset here so `Equipment Point` list page can point to a `organization-detail` url

saskatoon/harvest/filters.py
```diff
@@ -191,6 +195,35 @@ class Meta:
...
+            if 'is_equipment_point' not in data:
+                data['is_equipment_point'] = True
+        super(EquipmentPointFilter, self).__init__(data, *args, **kwargs)
+
+    equipment__type = filters.ModelChoiceFilter(
+        queryset=EquipmentType.objects.all(),
+        label=_("Available Equipment Type"),
+        help_text="",
+        required=False
+    )
```
2ynn, line 224: mmm I'm not sure I this approach very much, what I would rather we do is:
- filter the respective querysets in `beneficiary_list` and `equipment_point_list` so no unnecessary object is passed to the serializer
- override the Meta class when subclassing OrganizationFilter

saskatoon/harvest/api.py
```diff
@@ -197,28 +198,71 @@ def list(self, request, *args, **kwargs):
...
+                'title': _("New Organization")
+            }
+        })
+
+    @action(detail=False, methods=['get'])
+    def equipment_point_list(self, request, *args, **kwargs):
+        self.template_name = 'app/list_views/equipment_point/view.html'
+        self.filterset_class = EquipmentPointFilter
+
+        response = super(OrganizationViewset, self).list(request, *args, **kwargs)
```
2ynn, line 252: I think we should keep things more simple and create 2 separate `ListAPIView`s + 1 `DetailAPIView`. This way each list view can clearly define it's filter and it's initial queryset. Example:
https://github.com/LesFruitsDefendus/saskatoon-ng/blob/e8ac1756095791359eacff3e1207cd22002d67b4/saskatoon/harvest/api.py#L251

2ynn (changes requested):

saskatoon/harvest/api.py
oxgr: I imagined the queryset to be identical no? that was the main reason i switched gears and chose to keep it all on the same Viewset. if not, it wouldn't make much sense to have a filters like `is_equipment_point` for the Equipment Points list.

these two lists can still override the base classes `self.queryset`. not sure what the three separate views have over this, though it wouldn't be too hard to change

saskatoon/harvest/api.py
```diff
@@ -197,28 +198,71 @@ def list(self, request, *args, **kwargs):
...
-        return Response({'data': response.data,
-                         'filter': get_filter_context(self),
-                         'new': {'url': reverse_lazy('beneficiary-create'),
-                                 'title': _("New Organization")
-                                 }
+        return Response({'organization': response.data,
+                         'data': Equipment.objects.filter(owner_id=pk)
                          })
 
+    @action(detail=False, methods=['get'])
```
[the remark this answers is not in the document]
oxgr: routing is a subfolder of `organization`!

so the `beneficary_list()` method turns into `.com/organization/beneficiary-list` which i feel looks neat and informative

saskatoon/harvest/filters.py
oxgr: mentioned this below, but i thought we were set on keeping the querysets identical so we can play with filters on both tabs?

saskatoon/harvest/api.py
```diff
@@ -197,27 +201,85 @@ def list(self, request, *args, **kwargs):
...
+class BeneficiaryListView(LoginRequiredMixin, generics.ListAPIView):
+    """
+    List view for organizations that are beneficiaries.
+    """
+
+    permission_classes = [IsPickLeaderOrCoreOrAdmin]
+    queryset = Organization.objects.all().order_by('-actor_id')
+    serializer_class = OrganizationSerializer
+    filter_backends = (filters.DjangoFilterBackend,)
+    filterset_class = BeneficiaryFilter
```
2ynn, line 238: What was the problem with keeping the BeneficiaryViewset as is? I am open to renaming it `OrganizationViewset` but as we discussed I don't think it needs to be refactored:
- list view used for the "Beneficiary" tab (in reality it is more a `Organization` list page since we are actually passing all organization and filtering `is_beneficiary=true` by default
- detail view is used for the Organization detail page

and then a separate `EquipementPointListView(LoginRequiredMixin, generics.ListAPIView):` (as you did)

2ynn (changes requested):

saskatoon/harvest/api.py
oxgr: the same can be said for equipment points - the only difference being the filter and template

i thought we were separating them both into single generics because both list views are hierarchically equal - operating on the same queryset.

rundown of options was:
1. single viewset (org detail, ben list) + eq_point list -> confusing url routes
2. single viewset w/ custom actions for lists -> better urls, but over-engineered
3. separate generics for all -> like (2) with no inheritance

keeping the viewset as-is would make it confusing why the eq_point list is functionally similar with beneficiary, but not included within the viewset e.g. as a custom action

2ynn: > i thought we were separating them both into single generics because both list views are hierarchically equal - operating on the same queryset.

No I disagree, the queryset for the equipment point list is more restrictive, it should only show equipment points (i.e. `org.is_equipment_point=true` )

2ynn: > 1. single viewset (org detail, ben list) + eq_point list -> confusing relationships and url routes

what is confusing about this option exactly? Imo there doesn't need to be a relationship between the `OrganizationViewSet` and the `EquipmentPointListView`. Imo we are simply adding a new route to list equipment point and the rest was working fine (besides the questionable naming)

oxgr: reverted to an `Organization` viewset and kept the `EquipmentPointListView`

oxgr:
- `OrganizationViewset` has detail with a list view accessible via `Beneficiaries` button in menu
  - `list_views/beneficiary/` template directory renamed to `list_views/organization/` for consistency

- `EquipmentPointListView` added to `harvest/api.py`
  - made a check for Core or Admin role to show *New Organization* button for *Equipment Points* as it now redirects to admin panel
  - queryset filtered to only include equipment points

saskatoon/harvest/api.py
```diff
@@ -6,28 +6,31 @@
...
     @returns {dic} filters: filters template dictionary
     '''
     f = viewset.filterset_class(viewset.request.GET, viewset.queryset)
     dic = {'form': f.form}
     if any(field in viewset.request.GET for field in set(f.get_fields())):
-        dic['reset'] = reverse(viewset.basename + '-list')
+        if basename is not None:
+            dic['reset'] = reverse(basename + '-list')
+        else:
+            dic['reset'] = reverse(viewset.basename + '-list')
```
2ynn, line 33: NIT
```suggestion
         dic['reset'] = reverse("{}-list".format(
            basename if basename is not None else viewset.basename
        ))
```

saskatoon/harvest/api.py
```diff
@@ -197,27 +200,80 @@ def list(self, request, *args, **kwargs):
         return Response({'data': response.data})
 
 
-class BeneficiaryViewset(LoginRequiredMixin, viewsets.ModelViewSet):
-    """Beneficiary viewset"""
+class OrganizationViewset(LoginRequiredMixin, viewsets.ModelViewSet):
+    """Organization viewset - detail pages for beneficiaries and equipment points"""
```
2ynn, line 204:
```suggestion
    """Organization viewset"""
```

saskatoon/harvest/api.py
```diff
@@ -197,27 +200,80 @@ def list(self, request, *args, **kwargs):
...
 
     permission_classes = [IsPickLeaderOrCoreOrAdmin]
     queryset = Organization.objects.all().order_by('-actor_id')
-    serializer_class = BeneficiarySerializer
+    serializer_class = OrganizationSerializer
     filter_backends = (filters.DjangoFilterBackend,)
     filterset_class = OrganizationFilter
-    template_name = 'app/list_views/beneficiary/view.html'
+
+    def retrieve(self, request, format='html', pk=None):
```
2ynn, line 212: NIT
```suggestion
    def retrieve(self, request, format='html', pk=None):
    """Organization detail view -  shared by beneficiaries and equipment points."""
```

saskatoon/harvest/api.py
```diff
@@ -197,27 +200,80 @@ def list(self, request, *args, **kwargs):
...
+
+        # default request format is html:
+        return Response({
+            'organization': response.data,
+            'data': Equipment.objects.filter(owner_id=pk) 
+        })
 
     def list(self, request, *args, **kwargs):
-        response = super(BeneficiaryViewset, self).list(request, *args, **kwargs)
+        """Organization list view is accessible via the Beneficiaries menu button."""
```
2ynn, line 228: NIT
```suggestion
        """Organization list view - accessible via the Beneficiaries menu button."""
```

saskatoon/harvest/api.py
```diff
@@ -197,27 +200,80 @@ def list(self, request, *args, **kwargs):
...
+        })
+
+
+class EquipmentPointListView(LoginRequiredMixin, generics.ListAPIView):
+    """
+    List view for organizations that are equipment points.
+    """
+
+    permission_classes = [IsPickLeaderOrCoreOrAdmin]
+    queryset = Organization.objects.all().filter(is_equipment_point=True).order_by('-actor_id')
```
2ynn, line 250:
```suggestion
    queryset = Organization.objects.filter(is_equipment_point=True).order_by('-actor_id')
```

saskatoon/harvest/api.py
```diff
@@ -197,27 +200,80 @@ def list(self, request, *args, **kwargs):
...
+            return response
+
+        context = {
+            'data': response.data,
+            'filter': get_filter_context(self, 'equipment-point'),
+        }
+
+        # NOTE: Creation of organization that is an equipment point is currently restricted to the admin panel.
+        # The `New Organization` button should only appear for Core or Admin members.
+        # Change this if Equipment Point creation can be done with a conventional form.
```
2ynn, line 269: NIT
```suggestion
        # NOTE: Creation of a new Equipment Point is currently only supported in the admin panel
        # due to the Equipment inline form not having yet been implemented.  The `New Organization` 
        # button is restricted to Core or Admin members and simply links to the Admin creation form.
        # Change the `url`  once Equipment Point creation can be done with a conventional form.
```

saskatoon/harvest/filters.py
```diff
@@ -191,6 +195,16 @@ class Meta:
     )
 
 
+class EquipmentPointFilter(OrganizationFilter):
```
2ynn, line 198: subclassing `OrganizationFilter` means it will get the `is_equipment_point` field, which we don't need. Please create a new, independant filter by redefining the `Meta class`

saskatoon/harvest/forms.py
```diff
@@ -610,25 +610,24 @@ class Meta:
...
 class EquipmentForm(forms.ModelForm):
-    def clean(self):
-        cleaned_data = super(EquipmentForm, self).clean()
-        bool1 = bool(self.cleaned_data['property'])
-        bool2 = bool(self.cleaned_data['owner'])
-        if not (bool1 != bool2):
-            raise forms.ValidationError(
-                _('Fill in one of the two fields: property or owner.')
-            )
-        return cleaned_data
```
2ynn, line 621: please don't delete this, this is still used by `EquipmentAdmin` to make sure we don't fill both the `owner` and `property`

saskatoon/member/autocomplete.py
```diff
@@ -106,3 +106,21 @@ def get_queryset(self):
             qs = qs.filter(q0 | q1 | q2)
 
         return qs.distinct()
+
+
+class EquipmentPointAutocomplete(autocomplete.Select2QuerySetView):
+    """Organizations that are Equipment Points"""
```
2ynn, line 112: let's clarify we return an `Actor` queryset (and not `Organization`) as it is being used for the `owner` field

saskatoon/member/autocomplete.py
```diff
@@ -106,3 +106,21 @@ def get_queryset(self):
...
+
+    def get_queryset(self):
+        if not self.request.user.is_authenticated:
+            return Actor.objects.none()
+
+        f1 = Q(organization__isnull=False)
+        f2 = Q(organization__is_equipment_point=True)
+        qs = Actor.objects.filter(f1 & f2)
```
2ynn, line 120:
```suggestion
        qs = Actor.objects.filter(organization__isnull=False, organization__is_equipment_point=True)
```

saskatoon/member/autocomplete.py
```diff
@@ -106,3 +106,21 @@ def get_queryset(self):
...
+
+        if self.q:
+            q0 = Q(organization__civil_name__icontains=self.q)
+            qs = qs.filter(q0)
```
2ynn, line 124:
```suggestion
            qs = qs.filter(organization__civil_name__icontains=self.q)
```

saskatoon/member/urls.py
2ynn, line 1: where is the `EquipmentPointListView` route defined ?

2ynn (changes requested):

saskatoon/member/urls.py
oxgr: view in `harvest/api.py` and url in `harvest/urls.py` since i thought keeping it close to the `OrganizaitonViewset` made sense.

does that work? should it move to `member/`?

saskatoon/harvest/forms.py
oxgr: this `clean()` only makes sense if we keep the `property` field, which we decided to remove for the basic form.

if we want to keep this field in the admin form, should i then create a separate form for `EquipmentAdmin`?

I reckon the ultimate goal is to do an inline form since it will be edited from the org detail page.

oxgr: old `EquipmentForm` returned as `EquipmentAdminForm` in `harvest/admin.py`

saskatoon/harvest/filters.py
```diff
@@ -191,6 +195,32 @@ class Meta:
...
+        required=False
+    )
+
+    equipment__type = filters.ModelChoiceFilter(
+        queryset=EquipmentType.objects.all(),
+        label=_("Available Equipment Type"),
+        help_text="",
+        required=False
+    )
+
```
2ynn, line 222:
```suggestion
```

saskatoon/harvest/urls.py
```diff
@@ -9,11 +9,18 @@
...
 router.register('equipment', api.EquipmentViewset, 'equipment')
-router.register('beneficiary', api.BeneficiaryViewset, 'beneficiary')
+router.register('organization', api.OrganizationViewset, 'organization')
 router.register('community', api.CommunityViewset, 'community')
 router.register('participation', api.RequestForParticipationViewset, 'participation')
 
 urlpatterns = [
+
+    # READ VIEWS
+    path('equipment-points/',
```
2ynn, line 19:
```suggestion
    path('equipment-point/',
```
use singular, we will most likely add a `equipment-point/create/` and `equipment-point/update` down the road

2ynn (changes requested):

saskatoon/harvest/filters.py
```diff
@@ -191,6 +195,32 @@ class Meta:
...
 
 
+class EquipmentPointFilter(filters.FilterSet):
+
+    class Meta:
+        model = Organization
+        fields = [
+            'neighborhood',
+            'is_beneficiary',
+            'is_equipment_point',
```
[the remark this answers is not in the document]
oxgr: true. it got left behind from a copy. removed

2ynn (approved):

PR #425, opened 2024-11-18
Improve installation related docs and setup.py

phyllis-yo:
- Improve installation guide and readme files for fixtures and tests.
- Specify exact versions for Django and djangorestframework packages in setup.py to avoid confusion.

----------
## Type of change:
- [x] Documentation change.
----------
## What's Changed:
- docs
- setup.py
-------
## Checklist:


INSTALL.md
```diff
@@ -3,11 +3,16 @@
 Please follow each part of this documentation in order to run your own instance of Saskatoon.
 
 ## Overview
-1. [Install requirements](#requirements).
-2. [Prepare `.env` settings](#.env-settings).
-3. [Set up local database](#local-database).
-4. [Set up Django](#django-setup).
-5. [Load initial data (fixtures)](#loading-initial-data-(fixtures)).
+1. [Requirements](#requirements)
+2. [Install Saskatoon](#installation)
```
2ynn, line 7:
```suggestion
2. [Installation](#installation)
```

INSTALL.md
```diff
@@ -205,75 +231,44 @@ To import a *.sql* dump file into an **empty** database:
 $ mysql -u <user> -p <db_name> < dump_file.sql
 ```
 
-*Warning*: was only able to import a dump file into a database of the **same name**. To get the original database name from the dump file:
-```
-$ head -n 5 dump_file.sql
-```
+> WARNING: was only able to import a dump file into a database of the **same name**. To get the original database name from the dump file:
```
2ynn, line 234:
```suggestion
> WARNING: the database you are importing the dump file into should be *empty* and have the **same name** as the original database. To get the original database name from the dump file:
```

2ynn (commented):
looks great thanks @phyllis-yo !

Note some of the anchors in the Overview might need to be updated

2ynn:
> @2ynn Besides some clarifications on the content and improvement on the flow (hopefully the docs are more reader-friendly now), I made two changes that I want to explain a bit:
> 
>     1. I changed the command to create Python virtual environment to `python3 -m venv venv` instead of using 'virtualenv' package, because `venv` is a built-in package and is the recommended way of creating virtual environment by Python docs. Using 'venv' is more straight-forward in my opinion. 'virtualenv' might be slightly faster at runtime but I don't think that's our concern right now.
Ok no issue with using the built-in `venv`

>     2. I specified the exact versions of Django and djangorestframework in setup.py. The latest djangorestframework (3.15.2) is no longer compatible with Django 3.2.25, but just one minor down (3.15.1) still is. Listing the exact versions shall clarify potential confusion.

What kind of confusion did you run into?  If we decide to pin versions I think we should do it across the board, not just django and restframework

INSTALL.md
```diff
@@ -73,126 +72,153 @@ Please follow each part of this documentation in order to run your own instance
...
+```
+(venv)$ python3 saskatoon/manage.py migrate --skip-checks
+```
+
+In case you have an error similar to this one:
 ```
 django.db.utils.OperationalError: (1071, 'Specified key was too long; max key length is 767 bytes')
 ```
-then you'll need to set your database to UTF-8:
+then you need set your SQLite/MySQL database to UTF-8:
```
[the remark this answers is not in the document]
phyllis-yo: Agree.

INSTALL.md
phyllis-yo: Agree.

INSTALL.md
```diff
@@ -3,11 +3,16 @@
...
-4. [Set up Django](#django-setup).
-5. [Load initial data (fixtures)](#loading-initial-data-(fixtures)).
+1. [Requirements](#requirements)
+2. [Install Saskatoon](#installation)
+3. [Configuration](#configuration)
+   1. [Prepare `.env` settings](#.env-settings)
+   2. [Set up local database](#local-database)
+   3. [Set up Django](#django-setup)
+4. [Launch Saskatoon server](#launch-the-saskatoon-server-on-localhost)
+5. [Load sample data (fixtures)](#loading-sample-data-(fixtures))
```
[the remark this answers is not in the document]
phyllis-yo: Fixed. Thx.

INSTALL.md
```diff
@@ -3,11 +3,16 @@
 Please follow each part of this documentation in order to run your own instance of Saskatoon.
 
 ## Overview
-1. [Install requirements](#requirements).
-2. [Prepare `.env` settings](#.env-settings).
-3. [Set up local database](#local-database).
-4. [Set up Django](#django-setup).
-5. [Load initial data (fixtures)](#loading-initial-data-(fixtures)).
+1. [Requirements](#requirements)
```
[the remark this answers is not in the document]
phyllis-yo: Agree.

phyllis-yo:
> > I specified the exact versions of Django and djangorestframework in setup.py. The latest djangorestframework (3.15.2) is no longer compatible with Django 3.2.25, but just one minor down (3.15.1) still is. Listing the exact versions shall clarify potential confusion.
> 
> What kind of confusion did you run into? If we decide to pin versions I think we should do it across the board, not just django and restframework

@2ynn I wrongly installed `setup.py` with `python3 setup.py install` and Python installed 3 or 4 packages incompatible with Django 3.2.25 (it did warn though). It took me a while to figure out which package actually caused problem at runtime (which was djangorestframework as I mentioned). Thus I think it would be better just to explicitly pin versions in setup.py and we would know the local installation is always aligned with production.

setup.py
```diff
@@ -22,11 +22,10 @@
         "cffi>=1.14.0",
         "crispy-bootstrap4>=2022.1",
         "cryptography>=3.3.2",
-        "Django>=3.0.14,<4.0.0",
+        "Django==3.2.25",
         "django-geojson==3.0.0",
-        # "django-leaflet>=0.29.1",
         "django-rest-knox==4.1.0",
-        "djangorestframework>=3.11.0",
+        "djangorestframework==3.15.1",
```
2ynn, line 28:
```suggestion
        "djangorestframework>=3.11.0,<3.15.1",
```

setup.py
```diff
@@ -22,11 +22,10 @@
         "cffi>=1.14.0",
         "crispy-bootstrap4>=2022.1",
         "cryptography>=3.3.2",
-        "Django>=3.0.14,<4.0.0",
+        "Django==3.2.25",
```
2ynn, line 25:
```suggestion
        "Django>=3.0.14,<4.0.0",
```

setup.py
phyllis-yo: Agree

setup.py
```diff
@@ -22,10 +22,10 @@
         "cffi>=1.14.0",
```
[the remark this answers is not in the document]
2ynn: @phyllis-yo I agree with @tristanlatr that adding a `requirements.txt` file would be confusing. I would prefer if we called it `pip_freeze.txt` or something like that. As @oxgr mentioned it should act more as a "lock" file (provided we have enough discipline to update it when the prod server configuration changes) and should not be used to install requirements. Instead `setup.py` should remain the source of truth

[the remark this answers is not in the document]
phyllis-yo: @tristanlatr @2ynn Thanks for the advice. I agree that naming it `requirements.txt` could be too aggressive. Renamed it to `pip_freeze.txt` and rephrased the related content in `INSTALL.md`.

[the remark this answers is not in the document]
2ynn: thanks @phyllis-yo !!

INSTALL.md
```diff
@@ -3,11 +3,16 @@
...
 ## Overview
-1. [Install requirements](#requirements).
-2. [Prepare `.env` settings](#.env-settings).
-3. [Set up local database](#local-database).
-4. [Set up Django](#django-setup).
-5. [Load initial data (fixtures)](#loading-initial-data-(fixtures)).
+1. [Set up requirements](#requirements)
+2. [Install Saskatoon](#installation)
+3. [Configuration](#configuration)
+   1. [Prepare `.env` settings](#.env-settings)
```
2ynn, line 9:
```suggestion
   1. [Prepare `.env` settings](#env-settings)
```

2ynn (commented):
looks great to me, thanks a bunch!
just one broken anchor left i think

INSTALL.md
phyllis-yo: Ah, good catch! Thx.

2ynn (approved):

PR #427, opened 2024-11-23
chore: add warning for org.is_equipment_point boolean in admin form

2ynn:
When adding equipment to an organization in the `Organization` admin page, one can easily forget to check the `is_equipment_point` box. This PR adds a warning to try an alleviate these kinds of omissions.

Note this is not necessarily an omission as it could be possible for an organization to have equipment but that it is not currently open to reservations.


oxgr (approved):
warning comes out fine and it's good as-is, but i wonder if this info would be more useful while editing the form, perhaps as a colored warning text above/below the `Is Equipment Point` checkbox.

more work to do on-the-fly form validation though, so maybe another time.

2ynn:
thanks @oxgr I've improved the form and added help texts for more clarity

2ynn:
@oxgr this method doesn't work because the many-to-many relationship with equipment is only updated after the organization model is being save. For example:
- create a new organization with no equipment and leave the equipment_point box checked: the org is still marked as equipment point in the list display. Then reopen and resave the form: the box properly gets unchecked.
2. add new equipment but don't check the box: no warning is emitted. Then reopen and resave the form, the warning appears.

I will need to use signals for that.

saskatoon/member/admin.py
```diff
@@ -386,6 +426,21 @@ def contact(self, org):
...
 
+    def save_related(self, request, form, formsets, change):
+        """Ensure is_equipment_point box was not mistakenly left checked/unchecked"""
+
+        super().save_related(request, form, formsets, change)
+
+        org = Organization.objects.get(actor_id=form.instance.actor_id)
+        if not org.equipment.exists():
+            org.is_equipment_point = False
+            org.save()
```
oxgr, line 437: edge case, but would it make sense to allow setting `is_equipment_point` without having equipment?

e.g. if an org is found to be an equipment point, but nobody has yet to find out what equipment they actually have. the toggle would be useful for pick leaders to look into it, especially if the org a lot closer to the harvest than others.

2ynn: The current mechanism to "book" and equipment point is to store all of its equipment in the `equipment_reserved` field so if an organization has no equipment it cannot be booked in practice. I would much rather we create a dummy equipment type if needed

2ynn: I will add a warning message when `is_equipmen_point` is overriden on behalf of the user

Issue #414, opened 2024-10-22
Improve harvest start/end date UI


oxgr:
## Type of Issue: **Feature request**

Currently, the harvest details page and calendar pop-up show the start + end time of a harvest, but only the start date.

This becomes confusing if a harvest spans multiple days, especially if the same start time and end time were used.

## Suggestion

Include the end date in the harvest details page and calendar pop-up.

## Screenshots/ Code Snippets:

Calendar pop-up

Harvest details

Actual start/end date/time

## Related URLs:

- `https://saskatoon.lesfruitsdefendus.org/harvest/`
- `https://saskatoon.lesfruitsdefendus.org/calendar/`

2ynn:
I personally don't mind the datetime display as it is now, I think it's safe to assume that a harvest should start and end the same day. What I think we should do instead is:
- clarify a flow for fruit groups: historically fake harvests are created for this purpose so they can appear in the calendar. I think we should clarify those needs and have a model dedicated to fruit groups
- once this is done I think we could simply change the form validation to make sure start day = end day (or better have 1 date field + 2 time fields)

2ynn:
see also https://github.com/LesFruitsDefendus/saskatoon-ng/blob/7ad8c547b1e279cde8b8b4af7e34ec23afcaa073/saskatoon/sitebase/views.py#L162

2ynn:
- For harvests we should always validate `start_date == end_date`: forms could be simplified to add a single date field + 2 time fields
- Fruit groups should no longer rely on fake harvests
- core members should be able to see an ordered list of `fruits` associated with their estimated ripeness period (this could also be represented in the calendar but if so there should be a toggle to hide fruit groups for more clarity).
- the ripeness period for each fruit should be adjustable by core members. This would be reflected in all orphan harvests
- orphan harvests should not have any date associated to them. (the date should only be required for other statuses)
- a orphan harvest should clearly indicate the suggested date range based on the fruit ripeness period.
- if a pickleader adopts a harvest and sets it's date outside the suggested range they should receive a warning. We can eventually decide to raise a validation error

2ynn:
#419 should be addressed first

2ynn:
related to #495

PR #433, opened 2025-01-21
Unit tests: fixtures for models

phyllis-yo:
Add fixtures for models in unit tests
----------
## Type of change:
- [ ] Documentation change.

-------
## Checklist:
- [ ] My code follows the [style guidelines](https://github.com/LesFruitsDefendus/saskatoon-ng/blob/develop/CONTRIBUTING.md) of this project.


phyllis-yo:
@2ynn This is a prototype/demo/MVP for the stuff we discussed last time about creating fixtures for db in unit tests. Rather than a "real" pull-request (and thus I bypassed/oversimplified a lot of things), I just want to show you that it is viable and what the structure would look like.

To continue on, I need your inputs on what need to be tested (and not) in general, which would also help me and clean up/better structure the codes. Any thoughts?

saskatoon/unittests/test_harvest.py
```diff
@@ -0,0 +1,36 @@
...
+    return create_property(trees=[tree_apple, tree_pear])
+
+
+@pytest.fixture
+def harvest_apple(db, property_with_apple_pear: Property, tree_apple: TreeType) -> Harvest:
+    return create_harvest(harvest_property=property_with_apple_pear, harvest_trees=[tree_apple])
+
+
+def test_create_tree_type_apple(tree_apple: TreeType) -> None:
+    assert tree_apple.name == "apple"
```
2ynn, line 26: these kinds of actions like creating a new `TreeType` are usually done in the admin panel so I'm not sure how relevant unit tests are here. For example this test is simply testing `Model.objects.create`, which is a core Django functionality. I wouldn't want to add too much overhead if models have to be changed. In other words let's make sure we're testing our internal "business" logic and not purely django-related things

2ynn (commented):
thanks @phyllis-yo for kicking this off! 
I think a good starting point if you're looking for internal logic to be tested are things related to `start_date` / `end_date` 
For example here: https://github.com/LesFruitsDefendus/saskatoon-ng/blob/3b7bf15934788a7ab2ccb9cfdf99ae918e9c07ac/saskatoon/sitebase/views.py#L134

This issue might also provide more context: https://github.com/LesFruitsDefendus/saskatoon-ng/issues/414

Your tests could also expose the lacks we have in our validation errors. For example we shouldn't be able to create a harvest with `start_date` < `end_date`. I know we enforce some of these rules in the form validation but it might be worth making the models themselves more robust (i.e. adding validation steps in the overriden `save` methods for example). 
To be clear I'm not asking for you to go and add those validations but I think unit tests could expose these kinds of shortcomings.

2ynn:
@phyllis-yo more generally I think it would be better to try and reuse the fixtures we already have
(e.g. [TreeType](https://github.com/LesFruitsDefendus/saskatoon-ng/blob/master/saskatoon/fixtures/harvest-treetype.json) )
 and create new ones as necessary so objects don't need to be created manually. I believe there must be built-in ways to use fixture within tests:
https://docs.djangoproject.com/en/5.1/topics/db/fixtures/#how-to-use-a-fixture
```
class MyTestCase(TestCase):
    fixtures = ["fixture-label"]
```

2ynn:
> Also, I temporarily commented out some default values in the Property model, otherwise pytest complains when tearing down the test db since no State/City/Country models are created. IMO it should be fine to create a Property instance even when there is no State/City/Country instance existing, and thus it is a bug. Could you verify it?

Note if you change models new migration files need to be generated. The `default=1` value must have been introduced to handle migration errors. I think replacing `default=1` by `blank=True` would be acceptable here

2ynn:
> To continue on, I need your inputs on what need to be tested 

I will try and clarify/formalize this. In the mean time I would suggest trying to reproduce a typical harvest flow, i.e. roughly:
1. a pending property get created
2. the property gets validated
3. a new harvest gets created for this property ("orphan")
4. a pickleader adopts the harvest ("adopted")
5. a date / time is set for the harvest ("scheduled")
6. Requests for Participation (RfP) are made for this harvest
7. Some RfPs are accepted, some are declined
8. The harvest gets marked as "Ready", RfPs  are no longer allowed
9. The harvest is marked as "succeeded"
10. Fruits are weighted and registered with the harvest

For more details you can refer to the pickleader docs:
https://lesfruitsdefendus.readthedocs.io/en/latest/pre-pick-guide/life-cycle-of-a-pick-on-saskatoon.html

saskatoon/unittests/test_harvest.py
phyllis-yo: Agree. This test is just a demo to show that the fixtures work :p

phyllis-yo:
@2ynn Thanks for the input! Now I have a better idea on what to test and will start to design a testing plan accordingly.

I don't 100% agree on reusing the existing fixtures in the `fixtures` folder, but I think we will have more info to re-evaluate/discuss this once we have a solid testing plan.

Superpat:
Hi @phyllis-yo I think this branch has diverged a bit too much from main, but if you want to rebase it I'll be happy to take a look.

phyllis-yo:
> Hi @phyllis-yo I think this branch has diverged a bit too much from main, but if you want to rebase it I'll be happy to take a look.

Hi @Superpat Thanks for the reminder. This branch is more like a prototype/demo/MVP than a true feature branch. I think I will do a new branch (with the latest main) when getting back to this. You can safely ignore it for now.

Issue #429, opened 2024-11-29
add Equipment Point fixture


2ynn:
There should be at least one equipment point set up when running `saskatoon/fixtures/init` script

PR #454, opened 2025-03-14
Serializer e2e test

phyllis-yo:
- Add pytest fixture to load Django db fixtures.
- Add e2e test for serializers (harvest list, harvest detail, property list, property detail, organization list, organization detail, community list).
- Fix a missing-fields issue in PersonSerializer.

----------
## Type of change:
- [ ] Documentation change.
----------
## What's Changed:
- Add tests
- Fix serializer bug
-------
## Checklist:
- [ ] My code follows the [style guidelines](https://github.com/LesFruitsDefendus/saskatoon-ng/blob/develop/CONTRIBUTING.md) of this project.


phyllis-yo:
@2ynn This is the prototype for the serializer tests we discussed last time. Once it looks good, I can parameterize it to cover other urls like harvest detail, property, and etc.

Two things I want your comment:
1. Currently I dump everything including the pagination info to the json file for comparison. I think that's fine, but do let me know if you think the test should only compare the harvest list data.
2. I notice that the response of harvest detail (e.g. `/harvest/1/`) contains an 'about' entry which is unnecessary and could interference with the testing (its value looks like `"<django_quill.fields.FieldQuill object at 0x108423e20>"`). I think it should be removed.

saskatoon/unittests/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+                    'harvest-harvestyield', 'harvest-comment', 'harvest-requestforparticipation', 'sitebase-content']
+
+AuthUser = get_user_model()
+
+
+@pytest.fixture(scope="module")
+def django_db_setup_with_fixtures(django_db_setup, django_db_blocker):
+    with django_db_blocker.unblock():
+        for fixture_name in FIXTURE_SEQUENCE:
+            call_command('loaddata', FIXTURE_PATH.joinpath(f"{fixture_name}.json"))
```
2ynn, line 20: Note (not urgent, that's okay as is for now) : 

I wonder if the CI pipeline could export a single, gitignored `json` file containing all the objects since the fixtures `init` script is already ran here: https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/.github/workflows/tests.yml#L56

Moving forward this would avoid having to maintain both `FIXTURE_SEQUENCE` and  this one: https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/fixtures/init#L16

saskatoon/unittests/conftest.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+@pytest.fixture
+def client_core_user(client):
+    user = AuthUser.objects.create_user(email="test@user.com", password="password1234")
+    user.add_role('core')
```
2ynn, line 26: could we try and reuse the users that are loaded via the fixtures ?
https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/fixtures/member-authuser.json#L46

saskatoon/unittests/test_serializer.py
2ynn, line 1: nit: i would call this file `test_serializers.py`

2ynn (commented):
Thank @phyllis-yo that's great!

Are you sure you still see a `about` Quill field on `origin/main` or `origin/dev` branch ? I believe it should no longer be there :
https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/harvest/api.py#L81
https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/harvest/serializers.py#L280

saskatoon/unittests/conftest.py
2ynn: To clarify what I mean: after running `saskatoon/fixtures/init` the CI could run:
```
python3 saskatoon/manage.py dumpdata > test_fixture.json
```

And then in your tests you would simply need to run `loaddata` on this fixture

Alternatively I think the `init` script could run this automatically, not just in CI. However make sure the exported file is .gitignored as this has caused confusions in the past

saskatoon/unittests/test_serializer.py
phyllis-yo: Agree.

saskatoon/unittests/conftest.py
phyllis-yo: The Quill field I see is not in the harvest list view but the harvest detail view. See the attached json (the field is at the very end):
[harvest_detail.json](https://github.com/user-attachments/files/19253493/harvest_detail.json)

As for the duplication of fixture loading sequence, I like the idea of centralizing the sequence, but my concern is on the local dev environment: developers shall also be able to update/re-align the test fixtures locally. I propose to have a separate `init` script in the tests folder for testing to generate a `test_fixture.json` (rename the current `db.sqlite3`, load the fixtures in a new db, dump to `test_fixture.json`, revert db). What do you think?

saskatoon/unittests/conftest.py
phyllis-yo: I'd prefer to keep it separate so that this fixture can still be useful for tests don't rely on the django fixtures in `\fixtures\` folder.

2ynn: why not? I think using fixtures is better as you will load AuthUser/Person pairs, which is better than simply creating a AuthUser. Plus I have a feeling like you will end up creating more users with more roles. So why not relying on fixtures? These are here for test purposes

phyllis-yo: While I can see your point, I want to make the tests 'minimal' in the sense that they are setup in a way that just enough to test the specific functionalities accordingly. Hence, if a test fails, it is easy to pin-point where goes wrong.

On the other hand, I don't see the current implementation as a major issue. We can always change this fixture if later we find it is not enough when more tests are added.

2ynn: following a similar logic, if we used the same fixtures for all tests then when something goes wrong we have only a single place to look for (the fixtures) and things can be reproduced manually since the same objects can be loaded and investigated using the admin panel. Of course it's not a high cost decision but I'm pretty certain I will want to move to using fixtures in the future (at least for some tests) so this is not a zero cost decision. I understand wanting things to me "minimal", however testing with improperly configured users (e.g. a user with no person attached) doesn't seem appropriate. Further down the line your own "minimal" fixtures won't become as minimal and maintaining them will be more work than is necesary.

2ynn: I want to add that when we make changes to the models and add migrations we already have to maintain the fixtures to make sure they conform to the new models and they still load properly. In my view doing so should be enough for tests to still pass without having to maintain another set of manually configured fixtures.

saskatoon/unittests/conftest.py
2ynn: `about` is the harvest public announcement and is used in the harvest detail view:
https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/sitebase/templates/app/detail_views/harvest/about.html#L18

2ynn: > As for the duplication of fixture loading sequence, I like the idea of centralizing the sequence, but my concern is on the local dev environment: developers shall also be able to update/re-align the test fixtures locally. I propose to have a separate init script in the tests folder for testing to generate a test_fixture.json (rename the current db.sqlite3, call the init script in \fixtures\ folder to load the fixtures in a new db, dump to test_fixture.json, revert db). What do you think?

sure you can have a separate script whose role is solely to dump what is currently in the db into a `test_fixture.json` so it can be called alone when developing. However:
- this "script" is just a oneliner, developer who need to re-align test fixtures locally should already know about `manage.py dumpdata` 
- feel free to reuse the existing `dumpdata` script we already have (you can even remove the "Are you sure.." question and dump to another file than `saskatoon.json`, it's no longer useful)
- I would still call this script at the end of the current fixtures init script so that running `fixtures/init` sets everything up
- no need to rename/revert anything, just dump the current content

2ynn:
@phyllis-yo w.r.t [this comment](https://github.com/LesFruitsDefendus/saskatoon-ng/pull/454#discussion_r1996835283)

please review PR #455 , this will fix the `about` field serialization. Thanks for catching this!

screenshot_2025-03-15_09:36:46

saskatoon/unittests/conftest.py
phyllis-yo: Since we are still at an early stage in setting up the testing grid, I think we (at least I) still need to collect more data/evidence to solidify some decisions (that's why we build prototypes, right?), and long back-and-forth solely in theory may not be a wise use of our time. How about we revisit this after we have 10 tests?

phyllis-yo:
@2ynn I parameterize the test for most endpoints, except for `equipment-point` and the related detail view. The current fixtures don't include data for equipment points. Not sure if that's on purpose?

Along the way, I also fixed a small bug in PersonSerializer used in the endpoint `community`.

BTW, I notice that in the dashboard "Beneficiaries" tab has a filter for "Type". Is it still necessary since equipment point now have it own tab?

2ynn:
> @2ynn I parameterize the test for most endpoints, except for `equipment-point`, `participation` and their related detail views. The current fixtures don't include data for equipment points. Not sure if that's on purpose? `participation` is not currently exposed in the dashboard. Should it be included in the tests?

Yes we need to add fixtures for equipment-points.  #429 

> `participation` is not currently exposed in the dashboard. Should it be included in the tests?

yea I agree I don't think we need `api.RFPViewset` at the moment. Request For Participation are mainly important inside the Harvest Detail view.

> Along the way, I also fixed a small bug in PersonSerializer used in the endpoint `community`.

I suggest you rebase onto `origin/dev`, I think I have already updated this serializer in #448 

> BTW, I notice that in the dashboard "Beneficiaries" tab has a filter for "Type". Is it still necessary since equipment point 
now have it own tab?

Yes i'd like to keep it there for now. For example a Pick Leader might be interested in finding a Beneficiary Organization that also offer equipment

saskatoon/unittests/conftest.py
2ynn: I think that's okay since the output `.json` will be .gitignored. This way as you develop you can add more objects and run tests on them without modifying the init fixtures if needed. It's a normal flow for me to reset the test database regularly when playing with migrations

phyllis-yo: IMO it never hurts to have graceful prevention as human tends to forget, and it is the other way around that a developer should take the extra effort to modify the script as an explicit reminder for themselves, since testing without modifying the init fixtures may result in a failure in CI when they push.

Anyway, I think this is detail. The codes were already in the last commit and CI won't mind to execute two more small steps, unless you really hate them and want me to remove.

2ynn:
> Then I'd suggest I keep that line commented out and do a separate PR for #429 where I'd add fixtures and turn on the serializer test for equipment-point.

sure 

> The problem is, if I rebase onto `dev`, then I'd lose the fix you made in `main` a few days ago about the other serializer.

`origin/dev` now has that commit

saskatoon/unittests/init
```diff
@@ -0,0 +1,23 @@
...
+set -e
+
+# get script directory no matter where it's called from (doesn't work w/ symlinks)
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+cd $SCRIPT_DIR
+
+# rename the current db and restore it at the end of process
+if [ -e ../db.sqlite3 ]; then
+    mv ../db.sqlite3 ../db_temp.sqlite3
+fi
```
2ynn, line 15: I don't think you can assume that it's a sqlite3 file and it's not necessarily located in `saskatoon` folder
https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/env.template#L20

saskatoon/unittests/init
```diff
@@ -0,0 +1,23 @@
...
+
+../fixtures/init
```
2ynn, line 17: I don't understand why you want to run `fixtures/init` twice:

tests.yml
```
        saskatoon/fixtures/init
        saskatoon/unittests/init
```

saskatoon/unittests/conftest.py
2ynn: I am trying to keep things as simple as possible and I am not sure exactly what you are trying to prevent. 

>  in local environment if a developer added more entries in the db and call this new init script

why would they rerun the init script after adding more entries?

> since testing without modifying the init fixtures may result in a failure in CI when they push.

that's great, that's what CI is for


Since you were already planning on running the `fixtures/init` script twice, it looks to me like you should consider separating the DB that is used for development (as configured in `.env` file)  from another DB you generate specifically for tests. In this case your temporary test DB can be assumed to be a `db.sqlite3` file and can also be .gitignored

phyllis-yo:
@2ynn 
> > The problem is, if I rebase onto `dev`, then I'd lose the fix you made in `main` a few days ago about the other serializer.
> 
> `origin/dev` now has that commit

Wait, I can no longer see the fix from #448 in `dev`:
https://github.com/LesFruitsDefendus/saskatoon-ng/blob/9e15b00eb6fea1f16c4e5467016f73bc26598320/saskatoon/member/serializers.py#L84

Am I missing anything?

saskatoon/unittests/init
phyllis-yo: The first one is for the selenium which uses the real db (I don't think that's a good idea) and the second one is to create the `testing_db.json` for the tests using django testing db. Yea, I don't like this, either.

Personally I don't mind turning off the selenium tests since the page permission is now tested by `test_permission.py` (even more comprehensively).

2ynn:
@phyllis-yo I just merged `equipment_point` fixtures in `origin/dev`  (c.f. #429 )

I will also try and get #456 through ASAP as it brings many changes to models, fixtures and serializers.

Sorry you're dealing with such a moving target :/ If you'd rather try and merge your PR first to avoid loosing your work I suggest you revert to using the hardcoded fixture sequence you had initially and we can improve the way the test database is initialized in a separate PR.

Thanks again for your time, it's very much appreciated < 3

2ynn:
> @2ynn
> 
> > > The problem is, if I rebase onto `dev`, then I'd lose the fix you made in `main` a few days ago about the other serializer.
> > 
> > 
> > `origin/dev` now has that commit
> 
> Wait, I can no longer see the fix from #448 in `dev`:
> 
> https://github.com/LesFruitsDefendus/saskatoon-ng/blob/9e15b00eb6fea1f16c4e5467016f73bc26598320/saskatoon/member/serializers.py#L84
> 
> Am I missing anything?

gosh you're right :face_exhaling:, thanks for flagging! I went too quick when rebasing. #458 should fix this

phyllis-yo:
@2ynn 
> @phyllis-yo I just merged `equipment_point` fixtures in `origin/dev` (c.f. #429 )
> 
> I will also try and get #456 through ASAP as it brings many changes to models, fixtures and serializers.
> 
> Sorry you're dealing with such a moving target :/ If you'd rather try and merge your PR first to avoid loosing your work I suggest you revert to using the hardcoded fixture sequence you had initially and we can improve the way the test database is initialized in a separate PR.

No worries. The baseline files are generated by codes, so it won't take me long to update them after your changes. Just let me know when you are done.

And I agree that it will be better to just use the hardcoded fixture sequence for now. We can improve that part later separately (Also, I read the `newenvfile.py` for Selenium tests and see how it creates a separate db `sqlite3-testing.db`. I take back what I said in my previous comment).

saskatoon/unittests/init
2ynn: I still think it'd be great to unify test flows in some way, ideally by running the `fixtures/init` script only once and exporting the content of the selenium db into a `testing_db.json` file that can be reused for unittests.

Feel free to move/edit the `newenvfile.py` if that makes sense.

Btw, how about keeping only a single `tests` folder in `saskatoon` ? e.g:
tests/
    |__ selenium/
    |__ permissions/
    |__ serializers/
    
Another option would be to store unitttests along with each app (`harvest` , `member` , `sitebase` ): 
https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/harvest/tests.py

2ynn:
@phyllis-yo `origin/dev` should now have all the changes that used to be in `dev` + 455
In case you prefer to keep using main please approve this one once everything looks ok: https://github.com/LesFruitsDefendus/saskatoon-ng/pull/459

phyllis-yo:
@2ynn According to our discussion, I made the following changes:

- Roll back the fixture loading to base on hardcoded sequence for now (created an issue for improvement as well as revisiting the testing client fixture, since it is messier than we initially thought and deserves its own PR https://github.com/LesFruitsDefendus/saskatoon-ng/issues/462 )
- Add the test for endpoint `equipment-point`
- Change the base branch to `dev` so that your development may benefit from these tests.

phyllis-yo:
@2ynn Just a quick reminder: can I have your code review for the latest change (summarized in my previous comment)?

saskatoon/unittests/test_serializers.py
```diff
@@ -0,0 +1,36 @@
...
+@pytest.mark.parametrize("endpoint, baseline_filename", ENDPOINT_BASELINE_PAIRS)
+@pytest.mark.django_db
+def test_serializers(django_db_setup_with_fixtures, client_core_user, endpoint, baseline_filename):
+    response = client_core_user.get(get_json_url(endpoint))
+    assert response.status_code == 200
+
+    response_json = json.loads(response.content)
+    with open(BASELINES_PATH.joinpath(baseline_filename), 'r', encoding='utf-8') as baseline_file:
+        expected_harvest_list = json.load(baseline_file)
+        assert response_json == expected_harvest_list
```
2ynn, line 31:
```suggestion
        assert response_json == json.load(baseline_file)
```

2ynn (commented):
Thanks @phyllis-yo, I only have a minor comment

PR #455, opened 2025-03-15
fix: harvest public anouncement serialization

2ynn:
As discussed [here](https://github.com/LesFruitsDefendus/saskatoon-ng/pull/454#issuecomment-2725490770) the Harvest public anouncement (`about` field) was improperly serialized.


phyllis-yo (approved):
Looks good :)

PR #558, opened 2025-06-13
Stylesheet update for better UX on mobile as well as other minor HTML related changes

tristanlatr:
## What's Changed:

- Better experience on mobile:
   - not using fixed width for container division 
   - removing left and right margin for smaller screen so more content can fix on the screen
- Fix duplicated footer and move the footer inside the body element (as it should)
- Reduce size of landing page image (on large screens)  so it takes 50% of the width
- Inverse the position of the landing page image the landing page text so the links to the guides are more easy to find

--------

## Screenshots:
Capture d’écran, le 2025-06-12 à 21 16 01
Capture d’écran, le 2025-06-12 à 22 25 37

### DIFF
--- a/saskatoon/sitebase/static/css/style.css
+++ b/saskatoon/sitebase/static/css/style.css
@@ -62,7 +62,7 @@ body {
     background:#fff;
 }
 .main-body {
-    min-height: calc(100vh - 95px); /* wrt. footer height */
+    min-height: calc(100vh - 90px); /* wrt. footer height */
 }
 .floatleft {float:left}
 .floatright {float:right}
--- a/saskatoon/sitebase/templates/app/base/login.html
+++ b/saskatoon/sitebase/templates/app/base/login.html
@@ -28,12 +28,12 @@
             {% block body %}
                 <!-- content -->
             {% endblock body %}
-        </div>  
-    
+        </div>
+
         <footer>
             {% block footer %}
             {% include 'app/base/footer.html' %}
             {% endblock footer %}
-        </footer> 
+        </footer>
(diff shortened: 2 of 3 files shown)


2ynn:
thanks @tristanlatr this was definitely needed

I'm not sure if this layout was intended:
screenshot_2025-06-16_21:01:08

If so I would at least try to remove the padding on the left side of the image

2ynn:
looks good except the footer is no longer sticking to the bottom of the page, which was the problem we were originally trying to solve (#202) 


https://github.com/user-attachments/assets/1c2d2987-6ccc-4ba4-b2ab-a4e026779a6e

saskatoon/sitebase/static/css/style.css
```diff
@@ -56,19 +56,24 @@
...
 body {
-    min-height: 100vh;
     margin: 0;
     position: relative;
-    padding-bottom: 100px; /* W! footer height */
     box-sizing: border-box;
     background:#fff;
 }
+.main-body {
+    min-height: calc(100vh - 95px); /* wrt. footer height */
```
2ynn, line 65: how did you determine the `95px` value
screenshot_2025-06-17_21:22:26
 ?
I see some whitespace between the bottom of the footer and the bottom of the page. If I change this value for `90px` it disappears

2ynn: Looks like 95px was working when you had still `padding-top: 20px` instead of `padding-top: 15px` in the `footer-copyright-area` 
Either way if there's a formula let's add it as a comment

PR #611, opened 2025-10-21
Equipment point availability

### LINKED ISSUE #420: Determine Equipment Point availability
### Context

The `Harvest` model currently has a `equipment_reserved` _ManyToMany_ field because in the legacy version it used to be possible reserve one or more `Equipment` for a harvest. 
https://github.com/LesFruitsDefendus/saskatoon-ng/blob/3308f376fa89b210b1be4ba02264f411967d0568/saskatoon/harvest/models.py#L481

**I suggest we keep working with this existing field, even if as a first iteration we are only interested in booking a single _Equipment Point_**. This will allow us to add more granularity further down the road, e.g.:
- support multiple Harvests sharing the same Equipment Point
- support a single Harvest borrowing equipment from two different Equipment Points

### Definitions:

-  for a given `Harvest`,  _reserving_ an _Equipment Point_ means that **all** equipment registered for the `Organization` is booked at once, i.e. all items are added to the `equipment_reserved` list for this `Harvest`.
- an _Equipment Point_ is considered _unavailable_ for a given timeslot if at least one of its equipment is reserved by an _active_ Harvest whose _time range_ overlaps with the timeslot. 
- a Harvest is considered _active_ if its status is either _Scheduled_ or _Ready_ and it is not in the past. 
- the _time range_ during which a Harvest reserves an equipment point should be `start_date - buffer` -> `end_date + buffer`, where `buffer` is a constant time delta (e.g. 1hour) to give enough time for pickleaders to pickup/return the equipment before/after the harvest. 

### Tasks:

- [x] Create a utility function that lists all available equipment points for a given datetime range
- [x] Route the  [EquipmentPointAutocomplete](https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/member/autocomplete.py#L114) and expand it to allow filtering by datetime range. When passed a `start_date`/`end_date` the `equipment-point-autocomplete/` enpoint should return all Equipment Points that are _available_ for this datetime range.
- [ ] Update fixtures to have at least one Harvest with a reserved equipment point
- [ ] Add unit tests to thoroughly test the utility function
- [ ] The reservation buffer should be parameterized

Superpat:
Fixes #420
----------
## Type of change:
- [ ] Documentation change.
----------
## What's Changed:
Added a util function for listing available equipment points.

-------
## Checklist:


Superpat:
Hmm, strange that the unit tests failed. It seems like it's not using the right version of typeguard, but I made sure to put in the pip freeze.

saskatoon/member/autocomplete.py
```diff
@@ -30,26 +35,35 @@ def get_queryset(self):
...
         groups = Group.objects.filter(name__in=roles).values('id')
         return queryset.filter(groups__in=groups).distinct()
 
+    """
+    I (Patrick) could not get this method to typecheck properly, mypy does not recognize
+    AuthUser, it calls it AbstractBaseUser and the qs = self.get_roles_queryset line
+    causes since it's already initiated as a QuerySet[AbstractBaseUser], but I'm
+    not sure if the Q invocations after are used on qs in the case of self.q, so I've
+    left it untyped for now.
+    """
```
2ynn: does `-> Queryset[AbstractBaseUser]` work? If not no worries I don't think it's a big deal to leave it untyped

(Also no need to leave your name in comments, we can use `git blame`)

saskatoon/member/utils.py
```diff
@@ -14,3 +22,50 @@ def reset_password(user: AuthUser) -> str:
...
+          message='end must be later then start')
+@deal.pre(__valid_date_contract,
+          message='Substracting the buffer from the start date and '
+          'adding the buffer to the end date must result in valid dates')
+@deal.pre(lambda _: _.start.tzinfo is not None and _.end.tzinfo is not None,
+          message='start and end must be offset aware')
+@typechecked
+def available_equipment_points(
+    start: datetime,
+    end: datetime,
```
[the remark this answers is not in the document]
2ynn: Personally I don't think the utility function signature should be driven by how easily it is to test it but more by how convenient it is to use in our code base. However thinking this through it might actually come in handy in the future when we want to filter available equipment points for a given time range (regardless of existing harvest). So great I'd say let's keep start/end as is but let's set the default value for the buffer if that works: 
```
def available_equipment_points(
    start: datetime,
    end: datetime,
    buffer: timedelta = timedelta(hours=1)
) ->  QuerySet[Organization]:
```

I don't remember if `timedelta(1)` is a valid constant or can only be set at runtime so you might need to do:
```
```
def available_equipment_points(
    start: datetime,
    end: datetime,
    buffer: Option[timedelta] = None
) ->  QuerySet[Organization]:
   if buffer is None:
       buffer = DEFAULT_EQUIPMENT_POINT_BUFFER
   ...
```

and then I suggest we create a simple wrapper for convenience, e.g.:
```
def harvest_available_equipment_points(harvest: Harvest) -> QuerySet[Organization]:
```

If there's a way to avoid import cycle errors and add it as a method in the `Harvest` class then that'd be even better imo:
```
class Harvest(models.Model):
    """Harvest model"""

  ...
  def available_equipment_points(self) -> QuerySet[Organization]:
      return utils.available_equipment_points(harvest.start, havest.end)
 ```

saskatoon/unittests/member/test_utils.py
```diff
@@ -0,0 +1,159 @@
...
+    )
+
+    for case in cases:
+        case()
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize("harvest", Harvest.Status, indirect=True)
+def test_available_equipment_points_end_during(db, harvest, equipment) -> None:
+    """Check availability when it ends during another harvest"""
```
2ynn: This test also tests that equipment reserved by harvests that are not SCHEDULED or READY are ignored

Superpat: I updated all the comments to be more precise

saskatoon/member/utils.py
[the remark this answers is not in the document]
Superpat: I added the default value for buffer, but I didnt find a way to get around the circular dependencies so I could not add it to the class.

saskatoon/member/utils.py
```diff
@@ -14,3 +23,72 @@ def reset_password(user: AuthUser) -> str:
...
+    harvest: Optional[Harvest],
+    buffer: timedelta = timedelta(hours=1)
+) -> QuerySet[Organization]:
+    """List all available equipment points for a given datetime range"""
+    try:
+        """ A buffer gives pick leaders a bit of leeway in picking up and returning
+             the equipment, since some harvest sites can be further away """
```
2ynn: This syntax is reserved to docstrings. See [Block Comments](https://peps.python.org/pep-0008/#block-comments)
```suggestion
        # A buffer gives pick leaders a bit of leeway in picking up and returning
        # the equipment, since some harvest sites can be further away
```
Please fix throughout this PR

saskatoon/member/utils.py
```diff
@@ -14,3 +23,72 @@ def reset_password(user: AuthUser) -> str:
...
+            Harvest.Status.SCHEDULED, Harvest.Status.READY
+        ])
+
+        not_conflicting = (start_between | end_between) & is_active
+        """ We need to exclude the present harvest from the results
+              so it doesnt conflict with itself
+        """
+        id = harvest.pk if harvest is not None else None
+        filters = ~Q(harvest__pk=id) & (not_conflicting
+                                        if id is not None else not_conflicting)
```
2ynn:
```suggestion
        filters = ~Q(harvest__pk=id) & not_conflicting
```

2ynn (approved):
:rocket:

PR #614, opened 2025-11-11
Replace flake8 with ruff

Superpat:
Use [ruff](https://docs.astral.sh/ruff/) to format the python code (and preserving quotes) and replace flake8 with it for the linting step, it doesnt pop up any new issues and it's much faster.


Superpat:
@2ynn ruff is fast enough for both linting and formatting that it could be realistically placed in the pre-commit hook, let me know if you want me to do that.

I've had bad experiences with those in the past, but it's only a problem if the tool is slow.

2ynn:
> I've had bad experiences with those in the past, but it's only a problem if the tool is slow.

If it gets too slow then let's just add a CI check with instructions on how to run it

2ynn (approved):

PR #623, opened 2026-01-23
feat: add Equipment Point Map

Superpat:
Adds a web component that wraps leaflet and allows statically defining markers and popup markup.


2ynn (commented):
Great progress! A few comments:

1. I find the map too large compared to the table.
Image

2. I find the map too zoomed out, I'd suggest zooming in a bit more by default, e.g.: 

Image

3. I don't think "Contact" in "Contact {name}" should be italic and it is not currently translated:
Image

4. I suggest we indicate whether the equipment point is also a Beneficiary organization in the pop up:
screenshot_2026-01-30_14:15:59

saskatoon/sitebase/templates/app/list_views/base.html
```diff
@@ -44,7 +44,8 @@
                             </div>
                         </div>
                     </div>
-
+                    {% block map %}
```
2ynn: I wonder if placing the map underneath the table could look better ? Since we have only a handful of equipment points I suspect we will be able to see that there is a map available when scrolling. On the other hand with a map too large we may miss the table

saskatoon/sitebase/templates/app/list_views/organization/organizations.html
2ynn: Keep in mind that this template is also used by the Beneficiairies Tab.

As such, I think we need to also add icons for Beneficiary orgs that are not equipment points

Image

Superpat: Currently I show two icons side by side if they are both.

2ynn (commented):
I really like the possibility of switching from Map to Table. That said I think we need to style it better:

- right now it's very hard to know which one is active or not by looking at the button

Image
Image

- what I would suggest is try and create a "tab" effect, a bit like we have on the main menu
Image
__


Regarding the equipment Icons:
- When an org is both EP and beneficiary I'd squeeze the two icons a bit closer to each other: 
Image

- I'm not convinced by how much the icon size increases when we hover on it
- I'm wondering if the small modal could pop up when hovering on it without having to click on it? 
- Let's try and format the modal a bit cuter (e.g. a bit of transparency?, border?,  styling on text? )
- I think we should better incentivize users to visit the detail page for more info, (e.g. adding ✔️  vs ❌  to clearly indicate if the org has equipment/accepts fruits alongside `more details`  links). A lot of the important information will be stored in equipment/beneficiary descriptions so we need to make them clearly accessible somehow: 
screenshot_2026-02-05_18:17:03
Image

2ynn (approved):
Amazing 🔥

saskatoon/unittests/baselines/equipment-point_list.json
2ynn: What happened to upcoming reservations?

2ynn (changes requested):

saskatoon/unittests/baselines/organization_list.json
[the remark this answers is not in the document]
Superpat: I've reverted it since you said it was intentional.

2ynn (approved):

Issue #588, opened 2025-07-24
Feat: Public Homepage / About / FAQ


2ynn:
Public information about Les Fruits Défendus is currently hosted on [Santropol Roulant's website](https://santropolroulant.org/en/what-is-the-roulant/collectives/fruits-defendus/), which confuses people.

The goal is to get rid of the `saskatoon` subdomain (a name that also confuses people) and host a single unified website on `https://lesfruitsdefendus.org/`

### Public homepage

When accessing the website as a normal (anonymous) user:

- one or more pictures
- a short description of Les Fruits Défendus mission (+ link to the About page)
- a block explaining how to volunteer for a harvest (+ link to calendar)
- a block explaining how to register a new tree (+ link to public create form)
- a closing statement (+ link to the FAQ page)

### About / Contact page

- Values
- Different roles (tree owners, organizations, harvesters, core members)
- Contact information

### FAQ page

 An accordion list of questions and answers 

## CMS

All content should be editable in the admin panel via the Page Content WYSIWYG editor (besides pictures which can be hardcoded in templates for now, e.g. [home.jpg](https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/sitebase/templates/app/index.html#L46) )

Currently the public homepage relies on the [VOLUNTEER_HOME](https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/sitebase/models.py#L43) Page Content type, which can be edited from the [admin panel](https://saskatoon.lesfruitsdefendus.org/admin/sitebase/pagecontent/1/change/)

More types can be added to fit our needs and more than one content type can be used per page to keep things modular and well organized. For example the About page could load three content types: ABOUT_VALUES, ABOUT_ROLES and ABOUT_CONTACT.

The content must be loaded within [sitebase/views.py](https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/sitebase/views.py#L28) and passed to the `TemplateView` context

The FAQ will require a new model that Qynn will help setting up. This model should allow admins to:
- order question/answer pairs as desired
- enable/disable any question/answer pair

## Comments

- The header will have to be updated with links to `About` and `FAQ` pages
- Fixtures should be updated with temporary content
- The Pickleader homepage (`index.html` when user is logged in) can be left as is.

PR #638, opened 2026-01-31
feat: add FAQ page

2ynn:
partially addresses #588  & relies on #636 being merged

This PR adds a FAQ page:
- `FAQList` and `FAQItem` models allow adding/reordering question/answers from the admin panel
- initial fixtures reflecting https://santropolroulant.org/fr/quest-ce-que-le-roulant/des-collectifs/fruits-defendus/

screenshot_2026-01-30_20:05:47
screenshot_2026-01-30_20:32:42

### DIFF
--- a/saskatoon/harvest/locale/fr/LC_MESSAGES/django.po
+++ b/saskatoon/harvest/locale/fr/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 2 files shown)


saskatoon/sitebase/static/js/faq-list.js
Superpat: I'm not sure why you use javascript here, I feel like this code could be replaced with https://developer.mozilla.org/fr/docs/Web/HTML/Reference/Elements/details

Might have to a little bit of css tweaking though.

Superpat (changes requested):

Superpat:
There's a redis error in docker: `docker: Error response from daemon: client version 1.40 is too old. Minimum supported API version is 1.44, please upgrade your client to a newer version.`

PR #642, opened 2026-03-30
feat: improve auth.group permissions management

2ynn:
### Problems

1. There are discrepancies between the `auth.group` IDs between the dev environment and the production database 
2. doc/permissions.org is a maintenance burden and has fallen out of date
3. Changing permissions requires manual configuration through the admin panel

### What's changed

This PR formalize `auth.groups` that users with the 'core' and 'pickleader' roles belong to as code.

A `PERMISSIONS` dictionary has been added to each app to maintain users' permissions.

Running `make permissions`  will make sure that:
1. all models are reflected in their app's `permissions.py` file
2. The `doc/permissions.md` file is up to date 

To apply permissions to the current database:
```
python3 saskatoon/manage.py apply_permissions
```


saskatoon/sitebase/permissions.py
```diff
@@ -0,0 +1,33 @@
...
+Action = str
+Permissions = dict[ModelName, dict[Action, set[Role]]]
+
+PERMISSIONS: Permissions = {
+    "pagecontent": {
+        "add":    set(),
+        "change": set(),
+        "delete": set(),
+    },
+    "faqitem": {
```
Superpat: So does an empty set here still allow core members to modify faq items in the admin view ?

Superpat (approved):
Looks good to me

Issue #650, opened 2026-04-13
Past Reservations button broken


Superpat:
Past Reservations button in Equipment Point Details page does not work. It gives a 404 error.

Superpat:
It works locally, so I think prod does something different around url handling. The local deployment seems to rewrite the url to be `harvest/?` while in prod it stays `harvest?`

Superpat:
@2ynn let's keep this open until this hotfix is pushed into production.

Superpat:
This works now.

PR #651, opened 2026-04-13
Fix past reservations button

Superpat:
While it worked locally, prod requires url's to end with `/` before the query part.


Superpat:
fixes #650

2ynn (approved):

Issue #633, opened 2026-01-25
Chore: add admin action to bump maturity dates by 1 year


2ynn:
Orphan harvests get created based on TreeType's maturity dates. If these dates are in the past, orphan harvests won't be created properly. For convenience, there should be a way to bulk edit maturity dates from the admin panel

 
<img width="2297" height="419" alt="Image" src="https://github.com/user-attachments/assets/e9c2d5bf-8b3b-4862-ae26-8356ab48fdc9" />

2ynn:
No longer needed, year is dynamically updated in `property_create_orphans()`
```
            def update_year(d):
                return d.replace(year=datetime.today().year) if d is not None else None

            h = Harvest.objects.create(
                status=Harvest.Status.ORPHAN,
                property=property,
                start_date=update_year(to_datetime(t.maturity_start)),
                end_date=update_year(to_datetime(t.maturity_end)),
            )
            h.trees.add(t)
```

PR #655, opened 2026-04-23
Property map view

### LINKED ISSUE #643: Add interactive map to Property list view
Aimed primarly at core members to look for available properties.

- [ ] Icons should intuitively indicate the property's status (similarly to [harvest statuses](https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/sitebase/templatetags/harvest_status.py#L12), e.g.
  - pending: "#666",
  - validated: "#2da4f0"
  - authorized: "#8bc34a"
  - unauthorized: "#ff2079"
  - inactive: N.A. (filtered out)

<br>

- [ ] Title should be short address only, formatted as a link to the property detail page

- [ ] Pop-up should contain the following information:
  - name
  - address
  - status
  - available trees
  - number of successful harvests  (and/or date of last successful harvest ? )
  - button that links to detail page

Superpat:
Adds a map of properties along with some minor changes to the properties table view, fixes #643 

This pr also changes the way status colours are handled.


saskatoon/sitebase/locale/fr/LC_MESSAGES/django.po
```diff
@@ -1301,6 +1341,21 @@ msgstr "Connexion"
...
+
+#~ msgid "authorized"
+#~ msgstr "autorisée"
+
+#~ msgid "unauthorized"
+#~ msgstr "non autorisée"
+
+#~ msgid "inactive"
+#~ msgstr "inactive"
+
```
2ynn: Please review/resolve all the fuzzy entries, many are inaccurate

saskatoon/sitebase/static/js/map/leaflet.css
2ynn, line 1: this file might need some cleanup? unless you prefer keeping the comments in

saskatoon/sitebase/templates/app/list_views/property/data_row.html
```diff
@@ -32,41 +34,37 @@
...
                     {{tree.fruit_name_fr}}
                 {% endif %}
-                {{ tree.fruit_icon|default:"" }}
             </div>
         {% endfor %}
     </td>
     <td>
         {% if property.ladder_available %}
-            <i class="fa fa-check text-success"></i>
+            <i class="fa fa-check" style="color: #440bd4;"></i>
```
2ynn: I'd prefer if we can avoid inlining colors and keep them as centralized as possible, could you use the `saskatoon-map-success` class here instead ?

saskatoon/sitebase/templates/app/list_views/property/data_row.html
```diff
@@ -32,41 +34,37 @@
...
         {% else %}
-            <i class="fa fa-times text-danger"></i>
+            <i class="fa fa-times" style="color: #c01c28;"></i>
```
2ynn:
```suggestion
            <i class="fa fa-times saskatoon-map-danger"></i>
```

saskatoon/sitebase/templatetags/harvest_status.py
```diff
@@ -9,14 +9,40 @@ def color(harvest_status: str) -> str:
...
+            (Harvest.Status.ORPHAN, "#c01c28"),
+            (Harvest.Status.ADOPTED, "#000"),
+            (Harvest.Status.SCHEDULED, "#2da4f0"),
+            (Harvest.Status.READY, "#e8ad2b"),
+            (Harvest.Status.SUCCEEDED, "#440bd4"),
+            (Harvest.Status.CANCELLED, "#c01c28"),
         ]
-    }.get(harvest_status, "#d4c7f9")  # btn-default
+    }.get(harvest_status, "#666")  # btn-default
+
```
2ynn, line 25: Please revert these for now until we're set on the new colors, or at least keep the old values commented out

saskatoon/sitebase/static/js/map/leaflet.css
Superpat: This is a leaflet file, I needed to comment out some stuff that was too bothersome, but I'd rather have the least diff possible with the base file.

saskatoon/sitebase/templatetags/harvest_status.py
Superpat: I switched the colours for the ones in your diagram.

saskatoon/harvest/filters.py
```diff
@@ -169,7 +162,7 @@ def season_filter(self, queryset, name, year) -> QuerySet[Property]:
             return queryset
 
         harvests = Harvest.objects.filter(start_date__year=year)
-        return queryset.filter(harvests__in=harvests)
+        return queryset.filter(harvests__in=harvests).distinct()
```
Superpat, line 165: I added this because I saw duplicates in the results.

Superpat:
@2ynn @Maev-e this one is ready for a thorough review.

2ynn:
@Superpat please update PR description and link to issue

saskatoon/harvest/api.py
```diff
@@ -146,6 +146,47 @@ def partial_update(self, request, pk=None):
         return response
 
 
+class PropertyMapView(LoginRequiredMixin, generics.ListAPIView[Property]):
+    """List view for organizations that are equipment points."""
```
2ynn:
```suggestion
    """Map associated with the Property list view."""
```

saskatoon/harvest/admin.py
```diff
@@ -4,6 +4,9 @@
 from django.urls import reverse
 from django.utils.safestring import mark_safe
 from member.models import AuthUser
+from leaflet.admin import LeafletGeoAdminMixin  # pytype: disable=import-error
+
+from saskatoon.settings import DEFAULT_LEAFLET_TILE
```
2ynn, line 9: If you don't mind bring down the `member.models import AuthUser` import along with other internal deps

saskatoon/harvest/serializers.py
```diff
@@ -93,14 +93,19 @@ class Meta:
...
     trees = TreeTypeSerializer(many=True, read_only=True)
     owner = serializers.SerializerMethodField()
     pending_contact_name = serializers.ReadOnlyField()
     owner_type = serializers.SerializerMethodField()
     similar_properties = serializers.SerializerMethodField()
     needs_orphan = serializers.ReadOnlyField()
+    longitude = serializers.ReadOnlyField()
+    latitude = serializers.ReadOnlyField()
+    status = serializers.ReadOnlyField()
+    owner_name = serializers.ReadOnlyField()
```
2ynn: is `owner_name` needed? The `owner` field should give you all the info you need I think
```suggestion
```

saskatoon/harvest/serializers.py
```diff
@@ -144,17 +149,24 @@ class Meta:
...
+            'status',
         ]
 
+    status = serializers.ReadOnlyField()
     neighborhood = serializers.StringRelatedField(many=False)  # type: ignore
     # mypy says it should be a NeighborhoodSerializer
     trees = PropertyTreeTypeSerializer(many=True, read_only=True)
     harvests = PropertyListHarvestSerializer(many=True, read_only=True)
+    succeeded_harvests = PropertyHarvestSerializer(many=True, read_only=True)
+    next_harvests = PropertyHarvestSerializer(many=True, read_only=True)
```
2ynn: Why do we need to redefine these fields here? Aren't they inherited from `PropertySerializer` ?

saskatoon/harvest/views.py
```diff
@@ -536,11 +536,15 @@ def property_create_orphans(request, id):
     num_created = 0
     for t in property.trees.all():
         if not harvests.filter(trees__in=[t]).exists():
+
+            def update_year(d):
+                return d.replace(year=datetime.today().year) if d is not None else None
```
2ynn, line 552: Please move this function to `sitebase/utils.py` with other datetime helpers

saskatoon/saskatoon/settings.py
```diff
@@ -312,3 +312,12 @@ def get_drf_template_mode() -> str:
 
 # Equipment Point reservation buffer in hours
 DEFAULT_RESERVATION_BUFFER = 1
+
+DEFAULT_LEAFLET_TILE = (
```
2ynn, line 316:
```suggestion
# Maps
DEFAULT_LEAFLET_TILE = (
```

saskatoon/sitebase/locale/fr/LC_MESSAGES/django.po
```diff
@@ -1304,6 +1341,24 @@ msgstr "Connexion"
...
+#~ msgid "Ladder"
+#~ msgstr "Échelle"
+
```
2ynn:
```suggestion
```

saskatoon/sitebase/templates/app/detail_views/property/breadcomb.html
```diff
@@ -22,15 +22,17 @@ <h1>
                                         {{ address }}
                                     {% endif %}
                                     </h1>
+                                    {% with succeeded_harvests|last as last_harvest %}
```
2ynn: nit: please indent the block nested under `{% with %}`

saskatoon/sitebase/templates/app/list_views/property/data_row.html
```diff
@@ -32,41 +34,45 @@
...
         </ul>
     </td>
     <td>
-        <div class="text-nowrap">{{ property.last_succeeded_harvest_date|date:"Y-m-d g:i A" }}</div>
+        {% with property.succeeded_harvests|last as last_harvest %}
+            {% if last_harvest %}
+            <a data-placement="top" data-toggle="tooltip" href="{% url 'harvest-detail' last_harvest.id %}"
+                    title="#{{ last_harvest.id }}: {{ last_harvest.start_date }} -
+                                   {% if last_harvest.pick_leader %}{{ last_harvest.pick_leader.name }}{% else %}{% trans 'Orphan' %}{% endif %}"
+            >
```
2ynn: Could a template tag help here to make it easier to read and avoid repetition in the `next_harvest` title ?

saskatoon/harvest/serializers.py
```diff
@@ -93,14 +93,19 @@ class Meta:
     country = CountrySerializer(many=False, read_only=True)
     title = serializers.ReadOnlyField(source="__str__")
     harvests = PropertyHarvestSerializer(many=True, read_only=True)
-    last_succeeded_harvest_date = serializers.ReadOnlyField()
+    succeeded_harvests = PropertyHarvestSerializer(many=True, read_only=True)
+    next_harvests = PropertyHarvestSerializer(many=True, read_only=True)
```
2ynn: Why do we need to pass all of that extra context even though we are already serializing all property `harvests` ? Looking at `property_detail.json` baseline I think  we're passing a lot of ununsed data. At the minimum create more minimal serializers for these two fields so you can pass only what you need. That being said I'm wondering if the filtering based on dates and status cannot simply be done via template tags

saskatoon/harvest/serializers.py
Superpat: There's an owner field and a pending contact field, the owner_name field uses pending contact if owner isnt set yet.

2ynn: If the property is still `pending` then it should not have a harvest (i.e. by the time a harvest is created a property has been `validated`). So using pending contact is not relevant here. Please use `owner` field instead, no need for extra field

saskatoon/harvest/serializers.py
Superpat: I didnt program this with limiting verbosity of generated json in mind, instead my goal was to make writting templates as simple and flexible as possible. But I can see about changing it.

Still, really wish I could ripout django-rest though.

2ynn: It's not just about verbosity when reviewing baselines, the point is to not unnecessarily increase the payload (i.e. loading time for the paginated list view). This is why serializers like `HarvestListPropertySerializer` exist instead of on `PropertySerializer` for all routes

saskatoon/harvest/serializers.py
Superpat: good point

saskatoon/harvest/serializers.py
Superpat: same goes for the trees field, but I'll see if I can remove them.

Superpat: ok yeah those fields where not necessary, thanks

saskatoon/harvest/views.py
Superpat: My intuition would be to not create a public function unless it's used at least twice.

2ynn: Up to you. My goal was to have a single place where we handle datetime operations to  serve as documentation  and to make sure we do things uniformly. This one in particular is less of a big deal i agree since it doesn't rely on timezone but I'm pretty sure this function can become handy in the future e.g.   https://github.com/LesFruitsDefendus/saskatoon-ng/issues/633

2ynn:
@Superpat to reduce headaches while handling conflicts on  `.css` files I suggest you run this before rebasing on top of main:
```
(venv)$ python3 -m pip install cssbeautifier
(vent)$ venv/bin/css-beautify -rn --space-around-combinator saskatoon/sitebase/static/css/*.css
```

Then after you've rebased you can use `tox -e css-beautify` to lint CSS files

saskatoon/sitebase/templates/app/list_views/property/data_row.html
Superpat: Template tag was wrapping attributes in a tag, but a template filter allowed me to get just the attributes.

saskatoon/harvest/models.py
```diff
@@ -133,6 +133,12 @@ class Meta:
             '-id',
         ]
 
+    class Status(models.TextChoices, Enum):
+        INACTIVE = 'inactive', _("Inactive")
+        UNAUTHORIZED = 'unauthorized', _("Unauthorized")
+        PENDING = 'pending', _("Pending")
+        AUTHORIZED = 'authorized', _("Authorized")
```
2ynn, line 141: Sorry I didn't catch this earlier but our current logic relies on checking the nullable `self.authorized` in a tri-state: https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/sitebase/templates/app/detail_views/property/features.html#L43 : 
- `None` -> Not yet authorized
- `True` -> Authorized
- `False` -> Unauthorized

A property can be waiting for authorization for the current season even if it has already been validated. I'd suggest this instead: 
```suggestion
        PENDING = 'pending', _("Pending")
        VALIDATED = 'validated' , _("Validated")  # a.k.a. "Not yet Authorized"
        UNAUTHORIZED = 'unauthorized', _("Unauthorized")
        AUTHORIZED = 'authorized', _("Authorized")
        INACTIVE = 'inactive', _("Inactive")
```

saskatoon/harvest/models.py
```diff
@@ -449,6 +442,33 @@ def needs_orphan(self):
...
+    @property
+    def status(self):
+        if not self.is_active:
+            return Property.Status.INACTIVE
+
+        if self.pending:
+            return Property.Status.PENDING
+
+        if not self.authorized:
+            return Property.Status.UNAUTHORIZED
```
2ynn, line 472:
```suggestion
        if self.pending:
            return Property.Status.PENDING
            
        if self.authorized is None:
            return Property.Status.VALIDATED

        if not self.authorized:
            return Property.Status.UNAUTHORIZED
```

saskatoon/unittests/baselines/property_detail.json
```diff
@@ -406,7 +406,153 @@
...
+                    "property": "Michelle Delepeche at 67 rue de l'Oseille",
+                    "status": "succeeded",
+                    "status_display": "Succeeded",
+                    "start_date": "2025-08-06",
+                    "role": "volunteer",
+                    "rfp_status": "accepted"
+                }
+            ],
+            "organizations_as_contact": [],
+            "properties": []
```
2ynn: What data do you actually need for your `last_harvest` and `next_harvest`? Do we really need to serialize all the the next harvest's pickleader information including all their own led harvests ? Why can't you define a new serializer with less information ? Here is an example of a harvest serializer with less fields: https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/member/serializers.py#L44

Totally fine if extra serializer fields are more straightforward than template tags/filter but I'd appreciate if you could at least reduce the amount of data being passed by defining a new serializer for `last_harvest` and `next_harvest`. I understand it's not a big deal since it's not in a only a detail view but I'd rather we don't let it grow unnecessarily. And to be clear i'm not arguing the serialized data is minimal today and there are no unnecessary fields already passed to the property detailed paged. Feel free to skim down the existing `PropertyHarvestSerializer` if that's what the issue is and we're already passing data we don't need. IMO it's easier to not add extra data when designing a new feature than trying to trim it down further down the line when we're missing context.

saskatoon/harvest/locale/fr/LC_MESSAGES/django.po
```diff
@@ -235,76 +231,83 @@ msgstr ""
...
+#| "about frequency in additional comments at the bottom)"
 msgid ""
 "if not, please include info about frequency in additional comments at the "
 "bottom"
 msgstr ""
-"si non, merci de préciser leur fréquence de production dans les commentaires "
-"ci-dessous"
+"Mon arbre (mes arbres) ou vigne(s) produise(nt) des fruits annuellement "
+"(sinon, merci de préciser leur fréquence de production dans les Commentaires "
+"ci-dessous)"
```
2ynn, line 265: this fuzzied message seems wrong

saskatoon/harvest/locale/fr/LC_MESSAGES/django.po
```diff
@@ -235,76 +231,83 @@ msgstr ""
...
 
-#: harvest/forms.py:315
+#: harvest/forms.py:313
 msgid "I would lend my ladder for another harvest nearby"
 msgstr "Je peux prêter mon échelle lors d'une cueillette à proximité"
 
-#: harvest/forms.py:320
+#: harvest/forms.py:318
+#, fuzzy
+#| msgid "Produces fruits every year"
```
```suggestion
```

saskatoon/harvest/locale/fr/LC_MESSAGES/django.po
```diff
@@ -235,76 +231,83 @@ msgstr ""
...
-"ci-dessous"
+"Mon arbre (mes arbres) ou vigne(s) produise(nt) des fruits annuellement "
+"(sinon, merci de préciser leur fréquence de production dans les Commentaires "
+"ci-dessous)"
 
-#: harvest/forms.py:328
+#: harvest/forms.py:326
 msgid "Have you already registered your property in the past?"
-msgstr "Avez vous déjà inscrit votre propriété par le passé?"
+msgstr ""
```
2ynn: "Avez vous déjà inscrit votre propriété par le passé?"

saskatoon/harvest/locale/fr/LC_MESSAGES/django.po
```diff
@@ -361,58 +364,59 @@ msgstr ""
...
+#: harvest/forms.py:462
 msgid "Leave this field empty to publish harvest as soon as possible"
 msgstr "Laissez ce champ vide pour publier la récolte le plus tôt possible"
 
-#: harvest/forms.py:475 harvest/forms.py:560
+#: harvest/forms.py:473 harvest/forms.py:558
 msgid "Harvest must be confirmed before an equipment point can be booked"
-msgstr "La récolte doit être confirmée avant de pouvoir réserver un point d'équipment"
+msgstr ""
+"La récolte doit être confirmée avant de pouvoir réserver un point d'équipment"
```
```suggestion
msgstr ""La récolte doit être confirmée avant de pouvoir réserver un point d'équipment"
```

saskatoon/harvest/serializers.py
```diff
@@ -79,9 +79,17 @@ class Meta:
...
+            return None
+
+        data = PersonSerializer(harvest.pick_leader.person).data
+
+        return {
+            'actor_id': data['actor_id'],
+            'name': data['name'],
+            'phone': data['phone'],
+            'email': data['email'],
+        }
```
2ynn, line 92: Do you actually use the `phone` field ? 

```suggestion
        return PickLeaderSerializer(harvest.pick_leader).data
```

2ynn (approved):
screenshot_2026-05-15_21:19:35

Nothing blocking but I think we could improve the legend by:
- replacing red apple with red cross
- improving the icon order so  it matches the harvest flow, i.e.:
1. Orphan Harvest
2. Adopted Harvest
3. Scheduled Harvest
4. Ready Harvest
5. Succeeded harvest
6. Cancelled Harvest

1. Pending Property
2. Validated Property
3. Authorized Property
4. UnAuthorized Property


screenshot_2026-05-15_21:31:28

2ynn:
I find the spacing around `Owner` awkward:
screenshot_2026-05-15_21:34:31

PR #672, opened 2026-04-27
fix: prevent booking equipment while another harvest is using it

### LINKED ISSUE #670: Bug: Equipment points can be double booked if the harvest time changes
## Steps to Reproduce:
- Log in as pick leader
- Create two harvest - One in the afternoon, one in the morning
- On the afternoon harvest, reserve an equipment point 
- On the morning harvest, reserve the same equipment point
- Change the harvest time of the morning harvest to move it to the afternoon
- Notice that the same equipment point is booked twice 

## Expected Behavior:
- When changing the time of the harvest, the pick leader is prompted to select a new equipment point if there is a conflicting schedule 

## Current Behavior:
- Pick leaders can double book equipment points 

Image

Image

2ynn:
Fixes #670 

1. Logic did not account for the case where another harvests "surrounds" the requested time range:     `harvest_i_start`  <=  `requested_start`  < `requested_end`  <= `harvest_i_end`

2. The code responsible to prevent a harvest reservation to conflict with itself:
```
       # We need to exclude the present harvest from the results
        # so it doesnt conflict with itself
        id = getattr(harvest, "pk", None)
        not_itself = ~Q(harvest__pk=id)
```
made it impossible to detect conflicts with other harvest for the same equipment point because the reserved equipment was getting  excluded from the query (not simply the harvest <-> equipment relationship) so other harvests reserving it would not be part for the conflicting_orgs queryset

### DIFF
--- a/saskatoon/harvest/models.py
+++ b/saskatoon/harvest/models.py
@@ -534,7 +534,7 @@ class Harvest(models.Model):
 
     SEASON_CHOICES = [(y, y) for y in range(datetime.now().year, 2015, -1)]
 
-    ALLOWED_TO_RESERVE = [Status.READY, Status.SCHEDULED, Status.SUCCEEDED]
+    CAN_RESERVE_EQUIPMENT = [Status.READY, Status.SCHEDULED, Status.SUCCEEDED]
 
     status = models.CharField(
         choices=Status.choices,
--- a/saskatoon/harvest/models.py
+++ b/saskatoon/harvest/models.py
@@ -534,6 +534,8 @@ class Harvest(models.Model):
 
     SEASON_CHOICES = [(y, y) for y in range(datetime.now().year, 2015, -1)]
 
+    ALLOWED_TO_RESERVE = [Status.READY, Status.SCHEDULED, Status.SUCCEEDED]
+
     status = models.CharField(
         choices=Status.choices,
         max_length=20,
--- a/saskatoon/member/autocomplete.py
+++ b/saskatoon/member/autocomplete.py
@@ -4,7 +4,7 @@ from typeguard import typechecked
 
 from harvest.models import Harvest
 from member.models import AuthUser, Organization, Person, Actor, Neighborhood
-from member.utils import available_equipment_points
+from member.utils import get_available_equipment_points
 from saskatoon.autocomplete import Autocomplete
 from sitebase.utils import parse_naive_datetime
 
@@ -152,7 +152,7 @@ class EquipmentPointAutocomplete(Autocomplete):
         except (Harvest.DoesNotExist, ValueError):
             harvest = None
 
-        return available_equipment_points(start, end, harvest).distinct()
+        return get_available_equipment_points(start, end, harvest)
 
 
 @typechecked
--- a/saskatoon/unittests/member/test_utils.py
+++ b/saskatoon/unittests/member/test_utils.py
@@ -5,7 +5,7 @@ import hypothesis.strategies as st
 
 from django.conf import settings
 from sitebase.utils import parse_naive_datetime
-from member.utils import get_available_equipment_points
+from member.utils import is_equipment_point_available, get_available_equipment_points
 from harvest.models import Harvest, Equipment, EquipmentType
 from member.models import Organization, Country, State, City, Neighborhood
 
@@ -57,7 +57,7 @@ def equipment(db) -> Equipment:
 
 
 @pytest.mark.django_db
-def test_available_equipment_points_fuzz() -> None:
+def test_get_available_equipment_points_fuzz() -> None:
     start_strat = st.datetimes(
         min_value=datetime.min,
         max_value=datetime.max - timedelta(hours=1),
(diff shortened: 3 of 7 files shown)


saskatoon/member/autocomplete.py
```diff
@@ -152,7 +152,7 @@ def get_queryset(self) -> QuerySet[Organization]:
         except (Harvest.DoesNotExist, ValueError):
             harvest = None
 
-        return available_equipment_points(start, end, harvest).distinct()
+        return get_available_equipment_points(start, end, harvest)
```
Superpat, line 155: Probably wasn't the right place to check for distinct points, but I'm not seeing a `distinct()` in `get_available_equipment_points`. Are you sure we dont need it ?

saskatoon/unittests/member/test_utils.py
```diff
@@ -122,6 +122,28 @@ def test_available_equipment_points_start_during(db, harvest, equipment) -> None
         assert points.count() == 1
 
 
+@pytest.mark.django_db
+@pytest.mark.parametrize("harvest", Harvest.Status, indirect=True)
+def test_available_equipment_points_around(db, harvest, equipment) -> None:
+    """
+    If the requested equipement point conflicts with another harvest, booked by another
```
Superpat: Isn't the 'booked by another' superfluous here ?

saskatoon/unittests/member/test_utils.py
```diff
@@ -122,6 +122,28 @@ def test_available_equipment_points_start_during(db, harvest, equipment) -> None
...
+    there should only be an available equipment point if conflicting
+    harvests are not scheduled or ready
+    """
+
+    harvest.equipment_reserved.set([equipment])
+    points = get_available_equipment_points(
+        harvest.start_date - timedelta(minutes=30),
+        harvest.end_date + timedelta(minutes=30),
```
Superpat: @2ynn I'm surprised this case wasn't caught by start_during and end_during. 

But good catch @Maev-e

Superpat (changes requested):

saskatoon/member/autocomplete.py
2ynn: If `get_available_equipment_points()` could return duplicates then the `distinct()` should have been added inside the helper. But why would `get_available_equipment_points` return duplicates? 
```
    return Organization.objects.filter(is_equipment_point=True).exclude(
        pk__in=booked_equipment.values('owner')
    )
```
is only excluding a set of `pk`s, I have a hard time seeing how duplicate orgs can show in the returned queryset

saskatoon/unittests/member/test_utils.py
2ynn: Wdym? There are different scenarios (last one was missing): 
```
    start_between = Q(harvest__start_date__gte=start, harvest__start_date__lte=end)  # start <= si <= end
    end_between = Q(harvest__end_date__gte=start, harvest__end_date__lte=end)  # start <= ei <= end
    wrapping = Q(harvest__start_date__lte=start, harvest__end_date__gte=end)  # si <= start < end <= ei
```

` si <= start < end <= ei` is not covered by `start <= si <= end` and `start <= ei <= end`

`si = query_harvest_start`
`ei = query_harvest_end`
`start = requested_start`
`end = requested_end`

Though I admit I was surprised the edge case where start and end datetimes are _equal_ wasn't covered with the first check

2ynn:
@Maev-e if you have time to run a regression on this branch would be great

saskatoon/unittests/member/test_utils.py
Superpat: I just tried removing the wrapped case in the function (but keeping the new around test) and all tests are passing, so I'm not sure there was a problem on that level.

I'll try to replicate the problem in the ui now.

2ynn: Explain why you think the wrapped check needs to be removed? When testing locally in the form I did manage to fix the wrapping case and Maeve confirmed

Now regarding the test still passing could it be due to the fact that 30min falls into the 1hour buffer. I think the test harvest duration should be better set (ideally using a variable) so we can test this case better.

Superpat:
@2ynn I think I'm done with this. If @Maev-e doesnt find any further problems you can review and we can merge.

saskatoon/harvest/models.py
```diff
@@ -534,6 +534,8 @@ class Status(models.TextChoices, Enum):
 
     SEASON_CHOICES = [(y, y) for y in range(datetime.now().year, 2015, -1)]
 
+    ALLOWED_TO_RESERVE = [Status.READY, Status.SCHEDULED, Status.SUCCEEDED]
```
2ynn:
```suggestion
    CAN_RESERVE_EQUIPMENT = [Status.READY, Status.SCHEDULED, Status.SUCCEEDED]
```

saskatoon/unittests/member/test_utils.py
```diff
@@ -5,27 +5,23 @@
...
 from django.conf import settings
 from sitebase.utils import parse_naive_datetime
-from member.utils import available_equipment_points
+from member.utils import is_equipment_point_available, get_available_equipment_points
 from harvest.models import Harvest, Equipment, EquipmentType
 from member.models import Organization, Country, State, City, Neighborhood
 
+tzinfo = timezone(timedelta(hours=-5))
+start = datetime(2025, 3, 2, hour=12, minute=0, second=0, microsecond=0, tzinfo=tzinfo)
+end = datetime(2025, 3, 2, hour=18, minute=0, second=0, microsecond=0, tzinfo=tzinfo)
```
2ynn:
```suggestion
TZINFO = timezone(timedelta(hours=-5))
START = datetime(2025, 3, 2, hour=12, minute=0, second=0, microsecond=0, tzinfo=tzinfo)
END = datetime(2025, 3, 2, hour=18, minute=0, second=0, microsecond=0, tzinfo=tzinfo)
```

saskatoon/unittests/member/test_utils.py
```diff
@@ -61,187 +57,241 @@ def equipment(db) -> Equipment:
...
+def test_get_available_equipment_points_fuzz() -> None:
+    start_strat = st.datetimes(
         min_value=datetime.min,
-        max_value=datetime.max,
+        max_value=datetime.max - timedelta(hours=1),
         timezones=st.timezones(),
         allow_imaginary=True,
     )
 
+    end_strat = start_strat.flatmap(
```
2ynn, line 68: what does "strat" stand for ?

2ynn (commented):
@Superpat  just a few formatting nits

PR #698, opened 2026-05-17
Fix None in family names

### LINKED ISSUE #256: Family Name is None
While registering a user, if the family name is left blank, the user is displayed as <First_Name> None. It would be nice to display just the first name if the family name is None.

image

Superpat:
Allow blank strings

fixes https://github.com/LesFruitsDefendus/saskatoon-ng/issues/256


saskatoon/member/migrations/0021_alter_person_family_name.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 4.2.28 on 2026-05-17 12:19
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
Superpat, line 6: I'm not sure if I need to add an explicit migration that makes `None` become `""`, I think that might be handled, but I'll wait till @2ynn gives me some feedback on allowing empty strings.

2ynn: `None` -> `""` on `migrate` works

saskatoon/member/migrations/0021_alter_person_family_name.py
```diff
@@ -0,0 +1,19 @@
...
+    dependencies = [
+        ('member', '0020_alter_org_coordinates'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='person',
+            name='family_name',
+            field=models.CharField(
+                blank=True, default='', max_length=50, verbose_name='Family name'
```
2ynn:
```suggestion
                blank=True, default='', max_length=50, verbose_name='Last name'
```

2ynn (approved):

PR #699, opened 2026-05-17
chore: clean up member serializers

2ynn:
- rename `PersonOwnerSerializer`  -> `OwnerPersonSerializer` (person serializer)
- introduce `PickLeaderPersonSerializer` ( != `PickLeaderSerializer -> auth_user)

### DIFF
--- a/saskatoon/member/serializers.py
+++ b/saskatoon/member/serializers.py
@@ -121,6 +121,12 @@ class ContactPersonSerializer(PersonSerializer):
         fields = ['actor_id', 'roles', 'name', 'email', 'phone']
 
 
+class PickLeaderPersonSerializer(PersonSerializer):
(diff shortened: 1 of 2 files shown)


saskatoon/member/serializers.py
```diff
@@ -121,7 +121,13 @@ class Meta:
         fields = ['actor_id', 'roles', 'name', 'email', 'phone']
 
 
-class PersonOwnerSerializer(serializers.ModelSerializer[Person]):
+class PickLeaderPersonSerializer(PersonSerializer):
+    class Meta:
+        model = Person
+        fields = ['actor_id', 'name', 'email', 'phone']
```
2ynn, line 127: I don't think we're actually using `phone` atm but could be useful

PR #725, opened 2026-05-19
fix: Authorize button visibility + pending True in PublicPropertyForm

2ynn:
- Authorize button was not showing when property had `authorized=None`
- Send Email assumed recipient was `property.owner.person` 
- Set `pending: True` before calling clean() on PublicPropertyForm

### DIFF
--- a/saskatoon/harvest/forms.py
+++ b/saskatoon/harvest/forms.py
@@ -283,6 +283,7 @@ class PublicPropertyForm(forms.ModelForm[Property]):
     class Meta:
         model = Property
         fields = (
+            'pending',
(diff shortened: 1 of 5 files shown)


saskatoon/harvest/forms.py
```diff
@@ -435,6 +437,8 @@ class Meta:
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+
+        self.initial['pending'] = True
```
2ynn, line 441: This prevents an error when super().clean() gets called since `pending` was not specified as True:  

c.f #696 

Image

PR #730, opened 2026-05-23
Improve way harvests are shown in property list views

Superpat:
Last harvest and harvest list was redundant

### DIFF
--- a/saskatoon/sitebase/templatetags/harvest_status.py
+++ b/saskatoon/sitebase/templatetags/harvest_status.py
@@ -93,11 +93,8 @@ def is_ready_or_succeeded(status: str) -> bool:
 @register.filter
 @typechecked
 def next_harvest_date(harvest):
-    start = parse_naive_datetime(harvest['start_date'], DATE_INPUT_FORMATS[0])
-    end = parse_naive_datetime(harvest['end_date'], DATE_INPUT_FORMATS[0])
-
-    if start is not None and end is not None and start < end:
-        return start.strftime("%Y-%m") + ' ' + _('to') + ' ' + end.strftime("%Y-%m")
+    if harvest['date_range'] is not None:
+        return harvest['date_range']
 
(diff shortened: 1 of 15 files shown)


saskatoon/sitebase/templatetags/harvest_status.py
```diff
@@ -88,6 +90,49 @@ def is_ready_or_succeeded(status: str) -> bool:
...
+@typechecked
+def next_harvest_date(harvest):
+    start = parse_naive_datetime(harvest['start_date'], DATE_INPUT_FORMATS[0])
+    end = parse_naive_datetime(harvest['end_date'], DATE_INPUT_FORMATS[0])
+
+    if start is not None and end is not None and start < end:
+        return start.strftime("%Y-%m") + ' ' + _('to') + ' ' + end.strftime("%Y-%m")
+
+    return harvest['start_date']
+
```
2ynn: Why not making use of the existing `harvest['date_range']` field? If it is set display it, if not fall back to `start_date`.  Looks like this filter exists only to change `-` in `_('to')`,  i'm not sure it's necessary

2ynn (approved):

2ynn:
screenshot_2026-05-23_07:35:43

Maybe renaming the last 2 columns headers to "Harvested" and "Upcoming" would be shorter and clear enough?

2ynn:
screenshot_2026-05-23_07:39:15
We might be able to squeeze the icons a bit more horizontaly by reducing the padding on icons
screenshot_2026-05-23_07:40:27

saskatoon/sitebase/templatetags/harvest_status.py
Superpat: I simply didn't see it !

Superpat:
I can the reduce the right padding, but touching the left one screws with other bootstrap rules, I dont want to get into it too much.

PR #735, opened 2026-05-29
feat: add circular background to map icons

Superpat:
Was necessary since some colours lacked contrast.

Also removed some unecessary code around year handling since the harvest year filter handles it.

Also made the org map marker it's own file, will come in use later.


2ynn:
> Was necessary since some colours lacked contrast.
> 
> Also removed some unecessary code around year handling since the harvest year filter handles it.
> 
> Also made the org map marker it's own file, will come in use later.

@Superpat for PRs like these a couple screenshots would be nice

PR #749, opened 2026-07-10
Optimize properties map

Superpat:
Add htmx to load marker content dynamically and
calculate the marker icon / colour in the serializer so we can elide the harvest list from it.

The template filters have some cruft in them, but I cant clean it all up until I apply these
optimizations to the other maps and there's no use doing that until we know this
makes enough of a difference in real conditions.

### DIFF
--- a/saskatoon/harvest/api.py
+++ b/saskatoon/harvest/api.py
@@ -169,7 +169,9 @@ class PropertyViewset(LoginRequiredMixin, viewsets.ModelViewSet[Property]):
             }
         )
 
-    def marker(self, request, pk=None):
+    def map_marker_info(self, request, pk=None):
+        """Property details displayed in map pop-up window"""
+
         self.template_name = 'app/list_views/property/marker.html'
         property = self.get_object()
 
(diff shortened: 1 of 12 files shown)


saskatoon/harvest/api.py
```diff
@@ -168,6 +169,13 @@ def list(self, request, *args, **kwargs):
             }
         )
 
+    def marker(self, request, pk=None):
```
2ynn: it's not immediately clear what `marker` is used for, I'd suggest a more explicit name or at least adding a docstring
```suggestion
    def map_marker(self, request, pk=None):
```

2ynn (approved):

Superpat:
Thanks ! 


-------- Courriel initial --------
De : Qynn Schwaab ***@***.***>
Envoyé : 18 juillet 2026 13 h 14 min 01 s HAE
À : LesFruitsDefendus/saskatoon-ng ***@***.***>
Cc : Patrick Marchand ***@***.***>, Author ***@***.***>
Objet : Re: [LesFruitsDefendus/saskatoon-ng] Optimize properties map (PR #749)

Merged #749 into main.

PR #762, opened 2026-08-18
Feat/inline harvest comment creation/editing

JonGlaab:
## Type of change:
----------
## What's Changed:
- Harvest comments are now made directly on the harvest page without opening a new comment page.
- Comments can be edited in-line on the harvest page
- Admin can delete comments 
- Comment components were restructured to be more modular
- added an edited property to the comments model
- added logic to the serializer so it can properly label comments as edited or not
- added a button to make it clear if there are more comments to see

--------
## Screenshots:
1. basic view with new comment at top, inline edit box, and 'older comments' button
basicWithEdit

2. 'older comments' button removed at bottom of the list
bottomScroll

3. delete message
deleteMessage

--------
## Affected URLs:
e.g. `127.0.0.1:8000/harvest/<int>`


saskatoon/harvest/urls.py
```diff
@@ -77,6 +77,11 @@
         views.harvest_yield_delete,
         name='harvest-yield-delete',
     ),
+    path(
+        r'^comment/delete/<int:pk>/',
```
2ynn: > WARNINGS:
?: (2_0.W001) Your URL pattern '^comment/delete/<int:pk>/' [name='comment-delete'] has a route that contains '(?P<', begins with a '^', or ends with a '$'. This was likely an oversight when migrating to django.urls.path().

```suggestion
        r'comment/delete/<int:hid>/',
```

saskatoon/harvest/views.py
```diff
@@ -426,6 +407,24 @@ def get_success_url(self):
...
+    SuccessMessageMixin[CommentForm],
+    DeleteView[Comment, CommentForm],
+):
+    """Delete an existing comment."""
+
+    model = Comment
+    success_message = _("Comment deleted!")
+
+    def test_func(self) -> bool:
+        return self.get_object().author_id == self.request.user.id or is_admin(self.request.user)
```
2ynn: Only admins should be able to delete a comment.

```suggestion
    def test_func(self) -> bool:
        return is_admin(self.request.user)
```

saskatoon/sitebase/static/js/edit-comments.js
```diff
@@ -0,0 +1,33 @@
+document.addEventListener('DOMContentLoaded', function () {
+    document.addEventListener('click', function (e) {
```
2ynn: Please explain why you chose to add the `click` listener on the whole document instead of the particular elements you are interested in.  Example: https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/sitebase/static/js/list-view-tabs.js

saskatoon/sitebase/templates/app/detail_views/harvest/comments/comment_content.html
```diff
@@ -0,0 +1,63 @@
...
+                        class="btn btn-sm btn-link text-primary edit-comment-btn btn-comment-action"
+                        data-comment-id="{{ comment.id }}"
+                        title="{% trans 'Edit comment' %}">
+                    <i class="fa fa-pencil"></i>
+                </button>
+            {% endif %}
+            
+            {% if user|is_admin %}
+                {% translate "Are you sure you want to delete this comment?" as delete_confirm_msg %}
+                <form method="post"
```
2ynn:
```suggestion
                <form method="delete" 
```

saskatoon/sitebase/templates/app/detail_views/harvest/comments/view.html
```diff
@@ -0,0 +1,35 @@
...
+                                    {{ comment.author.email }}<br>
+                                    {% if comment.date_updated %}
+                                        {{ comment.date_updated }} ({% trans "edited" %})
+                                    {% else %}
+                                        {{ comment.date_created }}
+                                    {% endif %}
+                                </div>
+                            </div>
+
+                            {% include 'app/detail_views/harvest/comments/comment_content.html' with comment=comment %}
```
2ynn: nit: 
```suggestion
                            {% include 'app/detail_views/harvest/comments/comment_content.html' %}
```

2ynn (changes requested):
I think it's crucial we improve the visibility of the scroll bar. When there are too many comments posted it is absolutely not clear there are more below (see 1st screenshot). Ideally we would show the latest comments by default (i.e. scroll bar at bottom) and indicate clearly there are more comments. Ok to address in separate PR if needed

Image
Image

saskatoon/sitebase/static/js/edit-comments.js
JonGlaab: This was my oversight. It's been some time since I made an event listener and I used ai to generate this but I didn't catch the generalization. This version could help with comments that appear after the page is loaded but since comments currently refresh the page when added, I'll target the edit buttons directly

saskatoon/harvest/serializers.py
```diff
@@ -294,7 +294,11 @@ class Meta:
 
     author = PickLeaderSerializer(many=False, read_only=True)
     date_created = serializers.DateTimeField(format=r'%a %b %-d %Y %-I:%M %p')
-    date_updated = serializers.DateTimeField(format=r'%a %b %-d %Y %-I:%M %p')
+    date_updated = serializers.DateTimeField(format=r'%a %b %-d %Y %-I:%M %p', allow_null=True)
+    edited = serializers.SerializerMethodField()
+
+    def get_edited(self, obj: Comment) -> bool:
+        return obj.edited
```
2ynn: I don't think having a separate `edited` field is necessary here
I'd still prefer if you defined `date_updated` as `SerializerMethodField()` and gate it using `obj.edited` 

just FYI had you wanted to pass `edited` you could have simply used `ReadOnlyField()`

saskatoon/sitebase/locale/fr/LC_MESSAGES/django.po
```diff
@@ -602,26 +602,53 @@ msgstr "Oui"
...
+msgid "Edit comment"
+msgstr "Modifier le commentaire"
+
+#: sitebase/templates/app/detail_views/harvest/comments/comment_content.html:24
+msgid "Are you sure you want to delete this comment?"
+msgstr "Êtes-vous certain(e) de vouloir effacer ce commentaire?"
+
+#: sitebase/templates/app/detail_views/harvest/comments/comment_content.html:32
+msgid "Delete comment"
+msgstr "Effacer le Commentaire"
```
2ynn, line 615:
```suggestion
msgstr "Effacer le commentaire"
```

saskatoon/harvest/migrations/0025_alter_comment_options.py
2ynn, line 1: Please submit only a single migration file per app per PR

saskatoon/sitebase/locale/fr/LC_MESSAGES/django.po
```diff
@@ -1523,6 +1537,9 @@ msgstr ""
 "Les propriétaires n'ont pas encore indiqué qu'ils veulent participer cette "
 "saison, il est permis de les contacter pour confirmer leur accord"
 
+#~ msgid "New comment"
+#~ msgstr "Nouveau commentaire"
+
 #~ msgid "is a Montreal-based urban fruit harvesting collective hosted by"
```
2ynn: feel free to cleanup all commented lines at the bottom

saskatoon/fixtures/harvest-comment.json
```diff
@@ -29,7 +29,7 @@
     "author": 2,
     "content": "Fruits were frozen :(",
     "date_created": "2026-01-25T18:48:46.512Z",
-    "date_updated": "2026-01-25T20:17:39.268Z"
+    "date_updated": "2026-01-25T18:48:47.510Z"
```
2ynn, line 32: That's great, I would also edit another entry so `date_updated` - `date_created` sits right above the threshold

2ynn: Btw I did not realize that our baselines are only covering the harvest detail for pk=1 (`harvest-detail.json`)  Would be good to have at least 1 non-edited + 1 edited comment in harvest 1. Feel free to move existing comments to `"harvest": 1` instead

2ynn (approved):
great work!

PR #770, opened 2026-09-02
fix: header text stays in-line #756

### LINKED ISSUE #756: Bug[UI] table header text stacks vertically on larger viewports
## Steps to Reproduce:
    Launch and put the website in french
    Create a harvest
    Have at least one pending volunteer
    Possibly needs a comment or text in another field
    'Mis à jour le' will be stacked vertically instead of horizontal if the view is wide enough
    'nombre de cueilleur.euse.s' is also affected
    if the screen is downsized, the horizontal scroll-bar adjusts properly

## Expected Behavior:
    text should remain properly horizontal at all sizes or be two lines at most. 
    clipping the text is another possibility ("Mis à...")

## Current Behavior:
    text is distorted 

1. the distorted text
Image

2. the desired appearance
Image

## Related URLs:
`https://saskatoon.lesfruitsdefendus.org/harvest/<Number>`

JonGlaab:
Fixes #756
----------
## Type of change:
- [ ] Bug fix: 

----------
## What's Changed:
- added parameters to keep all table columns stable at different screen widths
- used descriptive styling to the css file to clean up the html
- kept 'nombre de cueilleur.euse.s' header wrapped to save space but added a Unicode space in '# Pickers' so it doesn't line-break
- added text to the Requester comment fixture to test with larger comments. Currently it displays max 100 chars but displays all on hover
- [UPDATED!] Found some icons that weren't loading correctly causing misalignment in some columns. Reinstated those in case that was desirable (picture below)
 
--------
## Screenshots:
1. full desktop
fullScreen

2. medium size screen
medScreen

3. phone screen
smallScreen

4. table with thumb and frown icons 
Screenshot 2026-09-15 at 9 19 35 AM

--------
## Affected URLs:
127.0.0.1:8000/harvest/<int>


saskatoon/sitebase/static/css/harvest-details.css
```diff
@@ -243,6 +243,52 @@
 
     th {
         font-weight: bold;
+        white-space: nowrap;
+    }
+
+
+    table th:nth-child(3) {
+        white-space: normal;
+    }
```
2ynn: why the need for 2 separate ` table th:nth-child(3)` entries?

saskatoon/sitebase/static/css/harvest-details.css
```diff
@@ -243,6 +243,52 @@
...
+
+    table th:nth-child(1) {
+        width: 14%;
+    }
```
2ynn: In this case I don't find it clearer to move the widths to a separate `.css` as it's not immediately clear which child id matches which header title.

saskatoon/sitebase/static/css/harvest-details.css
```diff
@@ -243,6 +243,52 @@
 
     th {
         font-weight: bold;
+        white-space: nowrap;
```
2ynn: IIUC this is the only line that affects the behavior, is that correct?

saskatoon/sitebase/static/css/harvest-details.css
JonGlaab: I could put these back into the html as is or I could give the columns class names in the html to make them explcit here

saskatoon/sitebase/static/css/harvest-details.css
JonGlaab: the last two columns were the only two without specified dimensions which seemed to be why the second to last was bouncing around. adding dimensions to keep them all consistent was my fix. 

but yes, I could remove all dimensions and this line basically does it. (I added this to keep the 'Chef.fe de cueillette' text in check also)

saskatoon/sitebase/templates/app/detail_views/harvest/data_table.html
```diff
@@ -13,13 +13,13 @@ <h2>{% trans "Pickers requests" %}</h2>
...
-                    <th style="width: 8%">{% trans "Rejections" %}</th>
-                    <th style="width: 9%">{% trans "Status" %}</th>
-                    <th style="width: 15%">{% trans "Pickleader note" %}</th>
+                    <th>{% trans "Requested on" %}</th>
+                    <th>{% trans "# Pickers" %}</th>
+                    <th>{% trans "Requester comment" %}</th>
+                    <th>{% trans "Experience" %}</th>
+                    <th>{% trans "Rejections" %}</th>
+                    <th>{% trans "Status" %}</th>
+                    <th>{% trans "Pickleader note" %}</th>
```
2ynn: How are we now limiting the column width? For example I think the "Requester comment" width should be capped so that even if the horizontal scroll bar appears for smaller screen sizes we it doesn't become much larger. Overall we want to make sure Action buttons are easily reachable

Image

saskatoon/sitebase/static/css/harvest-details.css
Superpat: yeah, table th.class and table td.class would probably be better.

saskatoon/sitebase/templates/app/detail_views/harvest/data_table.html
Superpat: Might be better to play with max-width / min-width rather then setting a width for each.

JonGlaab: This makes sense. On my limited testing I was under the impression there was a more global css affecting this table depending on screen size. I wasn't using a long enough comment to break it but max/min is what I'll try today

saskatoon/sitebase/locale/fr/LC_MESSAGES/django.po
```diff
@@ -653,40 +653,91 @@ msgstr "Oui"
...
+msgid "Comments"
+msgstr "Commentaires"
+
+#: sitebase/templates/app/detail_views/harvest/comments/view.html:20
+msgid "edited"
+msgstr "modifié"
+
+#: sitebase/templates/app/detail_views/harvest/comments/view.html:32
+msgid "Older comments"
+msgstr "Plus anciens"
```
2ynn, line 705: bad rebase ?

JonGlaab: The worst. changes in this PR moved a single translation one line down and then it kept fighting me over the date change on the PO files after running translate. Then it demanded another merge after the first PR went through today.

JonGlaab: The actual issue was translations getting broken by presence of '&nbsp;'. I tried to use '#&nbsp;Pickers' to keep that string unbroken while allowing the translation (nombre de cueilleur.euse.s) to wrap or word-break. I have switched the null char to a uncode friendly ' ' character that translate tolerates

saskatoon/sitebase/static/css/harvest-details.css
```diff
@@ -247,6 +246,59 @@
...
+    }
+
+    .col-date {
+        min-width: 10%;
+        white-space: nowrap;
+    }
+
+    th.col-number.lang-en::before {
+        content: "# ";
+    }
```
2ynn: What is this for? Is there an issue with `<th class="col-number">{% trans "# Pickers" %}</th>` ? 

```
#: sitebase/templates/app/detail_views/harvest/data_table.html:17
msgid "# Pickers"
msgstr "nombre de cueilleur.euse.s"
```

Should we simply have this instead?
```
#: sitebase/templates/app/detail_views/harvest/data_table.html:17
msgid "Num. of Pickers"
msgstr "N. de cueilleur.euse.s"
```

JonGlaab: This was an accident. I was trying to find a way to have '# Pickers' not wrap while french did wrap. I forgot to delete this but the issue was solved by using a unicode space char

2ynn (approved):

PR #774, opened 2026-09-06
Optimize equipment point map

Superpat:
Use htmx to dynamically load map marker info.


saskatoon/member/api.py
2ynn:
```suggestion
from rest_framework import viewsets
```

2ynn (approved):

Superpat:
thanks !

PR #780, opened 2026-09-10
Fix/long comments hidden, comment redirection and message correction 598

### LINKED ISSUE #598: [UI][CommunityTab][Comment] Comments on the community tab have no character limit and can break the UI when too long
## Steps to Reproduce:
- Go to the community tab
- Change a user's comment to make it extremely long
- Notice the column are broken after 

## Note
Most comments on the community tab are fine, however some can get very long, causing the specific user's line to be much bigger than the others. I think it might be worth it to have a maximum number of characters displayed there, with an option to expand to see the full comment. 

## Expected Behavior:
- Shown comments are limited to a certain number of character, and can be expanded to see the full comment

## Current Behavior:
- Comment have no character limit and can cause the column to break if they are too long

JonGlaab:
Fixes #598
----------
## Type of change:
- [ ] Bug fix 
----------
## What's Changed:
- added css and js to allow long comments to be truncated, added a button to expand them.
- added word-break to the user emails because a long email will also break containment here.
- removed 'required=False' from comment form to allow proper error messaging if a user bypasses browser level warnings
- added specialized mixins to add tags to comment form messages.
- tags allow component specific messages to be displayed near the component and not at the top.
- page refresh loads at the comment section after comment activity
- added better error handling in case a comment is too long. no longer crashes, names the mistake.
- however, overlong comment text is lost on refresh. I attempted to save it but perhaps this is a separate issue? (This one is already bloated)

--------
## Screenshots:
1. the original long email and long comment on mobile view
mobileBroken

2. mobile view after update (email may require further shortening)
mobileFix

3. desktop compressed comment
desktopSmall

4. desktop expanded comment
desktopLong

5. comment too long error
commentTooLong

6. placement of successful comment notification
afterNewComment

--------
## Affected URLs:
127.0.0.1:8000/harvest/<int>
127.0.0.1:8000/property/<int>

### DIFF
--- a/saskatoon/harvest/views.py
+++ b/saskatoon/harvest/views.py
@@ -400,6 +400,16 @@ class CommentCreateView(
             return reverse_lazy('property-detail', kwargs={'pk': pid})
         return reverse_lazy('home')
 
+    def form_invalid(self, form):
+        error_messages = []
+        for field, errors in form.errors.items():
+            for error in errors:
+                field_name = field.capitalize() if field != '__all__' else ''
+                error_messages.append(f"{field_name}: {error}" if field_name else error)
+
+        messages.error(self.request, " | ".join(error_messages))
+        return HttpResponseRedirect(self.get_success_url())
+
 
 class CommentUpdateView(
     LoginRequiredMixin,
@@ -431,6 +441,16 @@ class CommentUpdateView(
         elif self.object.property_id:
             return reverse_lazy('property-detail', kwargs={'pk': self.object.property_id})
         return reverse_lazy('home')
+
+    def form_invalid(self, form):
+        error_messages = []
+        for field, errors in form.errors.items():
+            for error in errors:
+                field_name = field.capitalize() if field != '__all__' else ''
+                error_messages.append(f"{field_name}: {error}" if field_name else error)
+
+        messages.error(self.request, " | ".join(error_messages))
+        return HttpResponseRedirect(self.get_success_url())
 
 
 class CommentDeleteView(
--- a/saskatoon/harvest/locale/fr/LC_MESSAGES/django.po
+++ b/saskatoon/harvest/locale/fr/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2026-09-07 19:35-0400\n"
+"POT-Creation-Date: 2026-09-10 18:17-0400\n"
 "PO-Revision-Date: 2025-06-13 11:05-0518\n"
 "Last-Translator: Anonymous User\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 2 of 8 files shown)


saskatoon/harvest/views.py
```diff
@@ -400,6 +400,16 @@ def get_success_url(self):
...
 
+    def form_invalid(self, form):
+        error_messages = []
+        for field, errors in form.errors.items():
+            for error in errors:
+                field_name = field.capitalize() if field != '__all__' else ''
+                error_messages.append(f"{field_name}: {error}" if field_name else error)
+
+        messages.error(self.request, " | ".join(error_messages))
+        return HttpResponseRedirect(self.get_success_url())
```
2ynn: What is preventing us from displaying form errors directly above the form like we already do for model forms?
See https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/sitebase/templates/app/forms/base.html#L18-L39 for example

saskatoon/harvest/views.py
```diff
@@ -432,6 +442,16 @@ def get_success_url(self):
...
+
```
2ynn: Would have preferred  not repeating exact same code across classes. Maybe a Mixin would have been justified here.

saskatoon/harvest/mixins.py
```diff
@@ -0,0 +1,17 @@
+from django.contrib import messages
+from django.http import HttpResponseRedirect
+from django.utils.translation import gettext_lazy as _
+
+
+class CommentFormInvalidMixin:
+    """Mixin to format form errors and redirect on invalid submissions."""
```
2ynn: What not keep a generic name so it could be reused elsewhere? Could the `content` fieldname be a class parameter? 
```suggestion
class FormInvalidMixin:
    """Mixin to format form errors and redirect on invalid submissions."""
```

JonGlaab: latest version tries to keep them generic but I'm not sure how many other places will want to use it because the mixins are how I obtained proper error messaging

saskatoon/harvest/views.py
```diff
@@ -395,22 +398,25 @@ def get_success_url(self):
         pid = self.kwargs.get('pid')
 
         if hid:
-            return reverse_lazy('harvest-detail', kwargs={'pk': hid})
+            url = reverse_lazy('harvest-detail', kwargs={'pk': hid})
         elif pid:
-            return reverse_lazy('property-detail', kwargs={'pk': pid})
-        return reverse_lazy('home')
+            url = reverse_lazy('property-detail', kwargs={'pk': pid})
+
```
2ynn, line 405: Should we raise in `else` case if none of `hid` and `pid` were provided?

2ynn (changes requested):
IIUC the whole form validation + `TaggedFormInvalidMixin` added complexity is solely to prevent the comment from being submitted with more characters than what `Comment`'s `content` field allows. IMO we should simply make sure the Text box doesn't allow more than the limit (and/or prevent the form to be submitted by disabling the button). ATM say I post a comment that is too long I essentially loose the whole comment content and i'm being redirected to an empty box with the error message. I don't think it's better than not being able to post it in the first place.

My suggestion is to simply start by setting `maxlength=500` on the `textarea` element. IMO it's pretty common and self explanatory

Image

saskatoon/harvest/views.py
JonGlaab: Yes we should, will add

PR #781, opened 2026-09-12
Fix/emails not normalized properly on user creation

JonGlaab:
Fixes #(no issue report)
----------
## Type of change:
- [ ] Bug fix 

----------
## What's Changed:
- Added get_auth_user helper to retrieve users case-insensitively using email__iexact
- Added create_auth_user utility function to sanitize and normalize email addresses before user creation
- updated user creation forms to use 'create_auth_user'
- updated user comparison forms to use 'get_auth_user'
- Implemented global email sanitization and lowercasing (.strip().lower()) directly within the AuthUser.save() method to act as a model-level backstop
- Overrode normalize_email on AuthUserManager to ensure lowercase handling across manager utilities

--------
## Screenshots:
1. Some emails before and after the normalization of the form
Screenshot 2026-09-12 at 3 03 06 PM

--------
## Affected URLs:
127.0.0.1:8000/participation/create/5/
--anywhere else where users were created with email

### DIFF
--- a/saskatoon/harvest/forms.py
+++ b/saskatoon/harvest/forms.py
@@ -15,7 +15,7 @@ from typing_extensions import Self
 
 from member.forms import validate_email
 from member.models import Organization, Person
-from member.utils import is_equipment_point_available, create_auth_user
+from member.utils import is_equipment_point_available, create_auth_user, get_auth_user
 from sitebase.models import Email, EmailType
 from sitebase.serializers import EmailRFPSerializer
 from sitebase.utils import is_quill_html_empty
@@ -266,7 +266,7 @@ class PropertyCreateForm(PropertyForm):
             auth_user = create_auth_user(email)
 
             person = auth_user.person
-            if not person: 
+            if not person:
                 person = Person.objects.create(
                     first_name=self.cleaned_data['owner_first_name'],
                     family_name=self.cleaned_data['owner_last_name'],
@@ -282,7 +282,7 @@ class PropertyCreateForm(PropertyForm):
                 )
                 auth_user.person = person
                 auth_user.save()
-                
+
             instance.owner = person
             instance.save()
 
--- a/saskatoon/harvest/forms.py
+++ b/saskatoon/harvest/forms.py
@@ -14,8 +14,8 @@ from typing import Any, Optional
 from typing_extensions import Self
 
 from member.forms import validate_email
-from member.models import AuthUser, Organization, Person
-from member.utils import is_equipment_point_available
+from member.models import Organization, Person
+from member.utils import is_equipment_point_available, create_auth_user
 from sitebase.models import Email, EmailType
 from sitebase.serializers import EmailRFPSerializer
 from sitebase.utils import is_quill_html_empty
@@ -63,10 +63,9 @@ class RFPForm(forms.ModelForm[RFP]):
 
     def clean_email(self):
         email = self.cleaned_data['email']
+        auth_user = get_auth_user(email)
 
-        if AuthUser.objects.filter(email=email).exists():
-            auth_user = AuthUser.objects.get(email=email)
-
+        if auth_user and auth_user.person:
             # check if a request with the same email already exists
             if RFP.objects.filter(person=auth_user.person, harvest_id=self.harvest.id).exists():
                 raise forms.ValidationError(_("You have already requested to join this pick."))
@@ -79,19 +78,17 @@ class RFPForm(forms.ModelForm[RFP]):
 
         # check if a user with the same email is already registered
         email = self.cleaned_data['email']
-        if AuthUser.objects.filter(email=email).exists():
-            auth_user = AuthUser.objects.get(email=email)
-            instance.person = auth_user.person
-        else:
-            instance.person = Person.objects.create(
+        auth_user = create_auth_user(email, ['volunteer'])
+
+        if not auth_user.person:
+            auth_user.person = Person.objects.create(
                 first_name=self.cleaned_data['first_name'],
                 family_name=self.cleaned_data['last_name'],
                 phone=self.cleaned_data['phone'],
             )
-            auth_user = AuthUser.objects.create(email=email, person=instance.person)
+            auth_user.save()
 
-        group, __ = Group.objects.get_or_create(name='volunteer')
-        auth_user.groups.add(group)
+        instance.person = auth_user.person
         instance.save()
 
         Email(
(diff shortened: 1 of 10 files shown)


saskatoon/member/models.py
```diff
@@ -132,6 +136,11 @@ def name(self):
             return self.person.name
         return None
 
+    def save(self, *args, **kwargs):
+        if self.email:
```
2ynn, line 140: in which scenario would `self.email` be unset?

saskatoon/member/models.py
```diff
@@ -132,6 +136,11 @@ def name(self):
...
+            self.email = self.__class__.objects.normalize_email(self.email).lower()
+        super().save(*args, **kwargs)
```
2ynn, line 142: - this save method belongs to the `AuthUser` class, **not** `AuthUserManager`:
- your implementation of `normalize_email()` already calls `.lower()`

```
    def save(self, *args, **kwargs):
        if self.email:
            self.email = type(self).objects.normalize_email(self.email)
        super().save(*args, **kwargs)
```

saskatoon/member/utils.py
```diff
@@ -97,3 +97,36 @@ def is_equipment_point_available(
...
+
+    if not email:
+        return None
+    normalized_email = AuthUser.objects.normalize_email(email)
+    user = AuthUser.objects.filter(email__iexact=normalized_email).first() # type: ignore[misc]
+
+    return cast(Optional[AuthUser], user)
+
+
+def create_auth_user(email: str, roles: List[Role]) -> AuthUser:
```
2ynn, line 113: - you shouldn't create a new `auth_user` without setting the `person` field
- name is misleading since you look for an existing user
```suggestion
def get_or_create_auth_user(person: Person, email: str, roles: List[Role]) -> AuthUser:
```

saskatoon/member/utils.py
```diff
@@ -97,3 +97,36 @@ def is_equipment_point_available(
...
+
+    if not email:
+        raise ValueError("Email is required.")
+
+    email = email.strip().lower()
+
+    # Check if they already exist (case-insensitively)
+    existing_user = get_auth_user(email)
+    if existing_user:
+        return existing_user
```
2ynn, line 124: If you return early here the roles won't get assigned

saskatoon/member/utils.py
```diff
@@ -97,3 +97,36 @@ def is_equipment_point_available(
...
+
+    # Cast to AuthUser for MyPy
+    user = cast(AuthUser, AuthUser.objects.create(email=email))  # type: ignore
```
2ynn, line 127: This looks wrong, should simply be:
```
user = AuthUser.objects.create(person=person, email=email)
```

saskatoon/member/utils.py
```diff
@@ -97,3 +97,36 @@ def is_equipment_point_available(
...
+    return cast(Optional[AuthUser], user)
+
+
+def create_auth_user(email: str, roles: List[Role]) -> AuthUser:
+    """Safely create or retrieve an AuthUser with full normalization."""
+
+    if not email:
+        raise ValueError("Email is required.")
+
+    email = email.strip().lower()
```
2ynn, line 119: Why don't you make use of the new `normalize_email()` function here?

saskatoon/harvest/forms.py
```diff
@@ -265,15 +262,26 @@ def clean(self):
...
+                auth_user.person = person
+                auth_user.save()
+            else:
+                person = Person.objects.create(
+                    first_name=self.cleaned_data['owner_first_name'],
+                    family_name=self.cleaned_data['owner_last_name'],
+                    phone=self.cleaned_data['owner_phone'],
+                )
+                auth_user.person = person
+                auth_user.save()
```
2ynn, line 284: This looks worse, the logic was right before, the only difference is that we want to call `normalize_email` before passing it, i.e.:
```
email = AuthUser.objects.normalize_email(self.cleaned_data['owner_email'])
auth_user = AuthUser.objects.create(email=email, person=person)
```
which could become `create_auth_user(person, email, ['owner']` with your helper

saskatoon/member/admin_forms.py
```diff
@@ -82,7 +83,13 @@ def __init__(self, *args, **kwargs):
...
+        email = data.get('email')
+
+        auth_user = get_auth_user(email)
+        person = None
+        if auth_user and auth_user.person:
+            person = auth_user.person
+
         if person:
             for key in ['first_name', 'family_name', 'phone']:
                 setattr(person, key, data.get(key))
```
2ynn, line 95:
```suggestion
        auth_user = get_auth_user(data.get('email'))

        if auth_user and auth_user.person:
            for key in ['first_name', 'family_name', 'phone']:
                setattr(auth_user.person, key, data.get(key))
```

saskatoon/member/admin_forms.py
```diff
@@ -114,21 +126,29 @@ def clean(self):
     def save_new_objects(self, commit=True):
         saved_instances = super().save_new_objects(commit)
         for person in [p for p in saved_instances if p is not None]:
-            email = self.get_emails().get(person.pk)
-            user, _ = AuthUser.objects.get_or_create(email=email, person=person)
+            email = self.get_emails().get(person.pk) or getattr(person, 'email', None)
```
2ynn, line 129: How would `getattr(person, 'email', None)` not return `None` since it's a proxy for `person.auth_user.email` ?

saskatoon/member/admin_forms.py
```diff
@@ -98,13 +105,18 @@ class PendingPickLeaderInlineAdminFormSet(
...
+            if (
+                f.instance.pk is not None
+                and f.cleaned_data
+                and not f.cleaned_data.get('DELETE', False)
+            ):
+                email = f.cleaned_data.get('email')
+                if email:
+                    # Normalize immediately so checks match database behavior
+                    emails[f.instance.pk] = AuthUser.objects.normalize_email(email).lower()
+        return emails
```
2ynn, line 119: I would rely on your helpers to handle normalization:
```
return dict(
            [
                (f.instance.pk, f.cleaned_data.get('email'))
                for f in self.forms
                if f.instance.pk is not None and f.cleaned_data and not f.cleaned_data.get('DELETE', False)
            ]
        )
```

saskatoon/member/admin_forms.py
```diff
@@ -114,21 +126,29 @@ def clean(self):
...
+            email = self.get_emails().get(person.pk) or getattr(person, 'email', None)
 
             # PickLeaders don't get assigned the 'pickleader' role until
             # they have read and agreed to the privacy policy
-            user.add_role('volunteer')
+            if email:
+                user = create_auth_user(email, ['volunteer'])
+                if not user.person:
+                    user.person = person
+                    user.save()
```
2ynn, line 137:
```suggestion
                user = get_or_create_auth_user(person, email, ['volunteer'])
```

2ynn (commented):
I think we should consider directly improving the `def create_user(self, email, password=None, **extra_fields)` method in `AuthUserManager` directly to prevent a `auth_user` being created without an associated `Person`

This PR should also come with  unit tests

saskatoon/member/utils.py
JonGlaab: It was failing tests and the errors seemed to point to it relating to an abstract user class that didn't like the extra email argument. so I tried to cast it, which still failed.

saskatoon/member/models.py
JonGlaab: In testing I found that AuthUserManager was still not being invoked with all user saves. Only changing the save in AuthUser caught all cases

Issue #484, opened 2025-03-30
Bug: Volunteers should be redirected to the calendar and error message should be displayed if they apply to the same harvest twice


Maev-e:
## Steps to Reproduce:
- Create a harvest and put it as "date scheduled"
- As a volunteer, go to the calendar
- Apply to the harvest
- Apply again to the harvest
- Notice you receive an error message within the form instead of being redirected to the calendar

## Expected Behavior:
Volunteer is redirected to the calendar and proper error message is shown, telling them they already applied

## Current Behavior:
Volunteer gets error messages within the form and is not redirected to the calendar

## Screenshots/ Code Snippets:

![Image](https://github.com/user-attachments/assets/63862333-f851-49a3-bcf3-c5e67966739c)

PR #783, opened 2026-09-14
Fix/duplicate volunteer error and redirect 483 484

### LINKED ISSUE #483: Bug: Wrong error message appears when pick leader add participants that already applied
## Steps to Reproduce:
- Create a harvest
- Add a participant using "New request"
- Add the same participant using "New request" 
- Notice the error message is not the right one 

## Expected Behavior:
Error message should be different than the one volunteers get when they apply twice to the same harvest

## Current Behavior:
Error message is the same as the one volunteers get when they apply to a new harvest 

## Note
Error message also appears twice on the form

Image

JonGlaab:
Fixes #483 and #484
----------
## Type of change:
- [ ] Bug fix
----------
## What's Changed:
- redundant error message removed from the request for participation form
- form now checks for three scenarios with individual error messages:
  1. the harvest leader tries to volunteer for their own harvest
  2. someone tries to add another person who is already signed up 
  3. a volunteer applies twice.
- in cases 1 and 2, the form refreshes with the appropriate error
- in case 3, if the volunteer applied from the calendar screen and not the harvest page, they are redirected back to the calendar with the appropriate error message

--------
## Screenshots:
1. case 1:
Screenshot 2026-09-14 at 5 59 15 PM

2. case 2:
doubleOther

3. case 3, redirected to calendar:
calendarRedirect

--------
## Affected URLs:
127.0.0.1:8000/participation/create/5/


saskatoon/harvest/locale/fr/LC_MESSAGES/django.po
```diff
@@ -162,108 +160,118 @@ msgstr "Non autorisée"
...
-#: harvest/forms.py:72
+#: harvest/forms.py:104
+msgid "The harvest pick leader cannot volunteer for their own pick."
+msgstr ""
+"Un.e chef.fe de cueillette ne peut pas se porter volontaire pour sa propre "
+"cueillette."
+
+#: harvest/forms.py:110
+msgid "This person has already been added to this pick."
+msgstr "Vous avez déjà demandé à rejoindre cette cueillette."
```
```suggestion
msgid "You have already submitted a request for this pick."
msgstr "Vous avez déjà demandé à rejoindre cette cueillette."
```

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,60 @@ class Meta:
     comment = forms.CharField(label=_("Comments"), required=False, widget=forms.widgets.Textarea())
 
     def __init__(self, *args, **kwargs):
+        self.request = kwargs.pop('request', None)
```
2ynn: What else do you need besides `request.user` ? 
```suggestion
        self.request_user = kwargs.pop('request_user', None)
```

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,60 @@ class Meta:
...
 
-        if AuthUser.objects.filter(email=email).exists():
-            auth_user = AuthUser.objects.get(email=email)
+        user_email = getattr(user, 'email', None)
+        leader_email = getattr(self.harvest.pick_leader, 'email', None)
+
+        if not user_email or not leader_email:
+            return False
 
+        return user_email.lower().strip() == leader_email.lower().strip()
```
2ynn:
```suggestion
    def is_harvest_leader(self) -> bool:
        return self.request_user == self.harvest.pick_leader
```

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,60 @@ class Meta:
...
+
+    def clean_email(self):
+        email = self.cleaned_data['email'].strip().lower()
+
+        auth_user = AuthUser.objects.filter(email__iexact=email).first()
+        if auth_user and auth_user.person:
```
2ynn: - stick to case sensitive emails for now
```suggestion
        email = self.cleaned_data['email'].strip()

        auth_user = AuthUser.objects.filter(email__iexact=email).first()
        if auth_user is not None and auth_user.person is not None:
```

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,60 @@ class Meta:
...
             # check if a request with the same email already exists
             if RFP.objects.filter(person=auth_user.person, harvest_id=self.harvest.id).exists():
```
2ynn: Note: this could also be written:

```suggestion
            if auth_user.person.requests.filter(harvest_id=self.harvest.id).exists():
```

https://github.com/LesFruitsDefendus/saskatoon-ng/blob/main/saskatoon/harvest/models.py#L811

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,60 @@ class Meta:
...
-                raise forms.ValidationError(_("You have already requested to join this pick."))
+                is_self_application = (
+                    self.request
+                    and self.request.user.is_authenticated
+                    and self.request.user.email.lower() == email
+                )
```
2ynn: we already know it's a self application with `if RFP.objects.filter(person=auth_user.person, harvest_id=self.harvest.id).exists():`

2ynn:
```suggestion
                is_self_application = self.request_user == auth_user
```

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,60 @@ class Meta:
...
+
+                leader_email = getattr(self.harvest.pick_leader, 'email', None)
+                target_email = getattr(auth_user, 'email', None)
+
+                is_target_harvest_leader = bool(
+                    leader_email
+                    and target_email
+                    and leader_email.lower().strip() == target_email.lower().strip()
+                )
```
2ynn: why aren't you using your `is_harvest_leader(self) -> bool` property here?

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,60 @@ class Meta:
...
             if RFP.objects.filter(person=auth_user.person, harvest_id=self.harvest.id).exists():
-                raise forms.ValidationError(_("You have already requested to join this pick."))
+                is_self_application = (
+                    self.request
+                    and self.request.user.is_authenticated
+                    and self.request.user.email.lower() == email
+                )
+
+                leader_email = getattr(self.harvest.pick_leader, 'email', None)
+                target_email = getattr(auth_user, 'email', None)
```
2ynn: why `geattr`? 
```suggestion
                leader_email = self.harvest.pick_leader.email
                target_email = auth_user.email
```

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,60 @@ class Meta:
...
+
+                is_target_harvest_leader = bool(
+                    leader_email
+                    and target_email
+                    and leader_email.lower().strip() == target_email.lower().strip()
+                )
+
+                if self.is_harvest_leader and is_self_application:
+                    raise forms.ValidationError(
+                        _("The harvest pick leader cannot volunteer for their own pick.")
```
2ynn: Since it's the pick leader that requests for themselves, why not using "*You* cannot volunteer for your own pick" ?

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,60 @@ class Meta:
...
+                    )
+                elif (
+                    self.is_harvest_leader or is_target_harvest_leader
+                ) and not is_self_application:
+                    raise forms.ValidationError(
+                        _("This person has already been added to this pick.")
+                    )
```
2ynn: why does it matter whether the requester is the pick leader here? The same error should be raised either way.

Conversely, if the target is the harvest leader `is_target_harvest_leader = True` then this is where we should say "The harvest pick leader cannot volunteer for their own pick."

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,60 @@ class Meta:
...
+                else:
+                    self.is_volunteer_duplicate = True
```
2ynn: what's this flag for?

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,60 @@ class Meta:
...
     def __init__(self, *args, **kwargs):
+        self.request = kwargs.pop('request', None)
         if 'harvest' in kwargs:
             self.harvest = kwargs.pop('harvest')
         super().__init__(*args, **kwargs)
 
-    def clean_email(self):
-        email = self.cleaned_data['email']
+    @property
+    def is_harvest_leader(self) -> bool:
```
2ynn:
```suggestion
    def is_requester_harvest_leader(self) -> bool:
```

saskatoon/harvest/forms.py
2ynn:
```suggestion
                is_target_harvest_leader = auth_user == self.harvest.pick_leader
```

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,60 @@ class Meta:
...
+
+        if not user_email or not leader_email:
+            return False
 
+        return user_email.lower().strip() == leader_email.lower().strip()
+
+    def clean_email(self):
+        email = self.cleaned_data['email'].strip().lower()
+
+        auth_user = AuthUser.objects.filter(email__iexact=email).first()
```
2ynn: calling it `target_user` may help reasoning here since you have `is_target_harvest_leader`

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,60 @@ class Meta:
...
+                    raise forms.ValidationError(
+                        _("The harvest pick leader cannot volunteer for their own pick.")
+                    )
+                elif (
+                    self.is_harvest_leader or is_target_harvest_leader
+                ) and not is_self_application:
+                    raise forms.ValidationError(
+                        _("This person has already been added to this pick.")
+                    )
+                else:
```
2ynn: I would have started with this case:
```
if not self.request_user.is_authenticated:
   raise forms.ValidationError(_("You have already requested to join this pick."))
else if...
```

2ynn (changes requested):

saskatoon/harvest/forms.py
JonGlaab: This longer block validates against anonymous users applying to a harvest while not signed in. But It can be simplified to 
'''
@property
def is_harvest_leader(self) -> bool:
        if not self.request_user or not getattr(self.request_user, 'email', None):
            return False
'''

saskatoon/harvest/forms.py
JonGlaab: to help send the error message to the calendar view to be displayed on redirect

saskatoon/harvest/forms.py
JonGlaab: changed here and wherever function is called

saskatoon/harvest/locale/fr/LC_MESSAGES/django.po
JonGlaab: updated both cases to match this phrasing:

msgid "This person has already submitted a request for this pick."
msgstr "Cette personne a déjà demandé à rejoindre cette cueillette."

msgid "You have already submitted a request for this pick."
msgstr "Vous avez déjà demandé à rejoindre cette cueillette."

saskatoon/harvest/forms.py
JonGlaab: updated!

saskatoon/harvest/forms.py
JonGlaab: I'm checking for email explicitly here bc I had a weird issue where the pick@leader.com pk wasn't registering as the same pk from the harvest leader (who was Pick Leader). I found email to work better for comparison

saskatoon/harvest/locale/fr/LC_MESSAGES/django.po
```diff
@@ -162,108 +160,118 @@ msgstr "Non autorisée"
...
 msgid "Comments"
 msgstr "Commentaires"
 
-#: harvest/forms.py:72
-msgid "You have already requested to join this pick."
+#: harvest/forms.py:89
+msgid "The harvest pick leader cannot volunteer for their own pick."
+msgstr ""
+"Un.e chef.fe de cueillette ne peut pas se porter volontaire pour sa propre "
+"cueillette."
```
```suggestion
msgid "A pick leader cannot volunteer for their own harvest."
msgstr ""
"Un.e chef.fe de cueillette ne peut pas se porter bénévole pour sa propre "
"cueillette."
```

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,48 @@ class Meta:
...
 
     def __init__(self, *args, **kwargs):
+        self.request_user = kwargs.pop('request_user', None)
         if 'harvest' in kwargs:
             self.harvest = kwargs.pop('harvest')
         super().__init__(*args, **kwargs)
 
+    @property
+    def is_requester_harvest_leader(self) -> bool:
+        if not self.request_user or not getattr(self.request_user, 'email', None):
```
2ynn:
```suggestion
        if self.request_user is None or self.request_user.is_anonymous:
```

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,48 @@ class Meta:
...
+            return False
+
+        return self.request_user.email == self.harvest.pick_leader.email
```
2ynn:
```suggestion
        return self.request_user == self.harvest.pick_leader
```

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,48 @@ class Meta:
...
+            if is_target_harvest_leader:
+                raise forms.ValidationError(
+                    _("The harvest pick leader cannot volunteer for their own pick.")
+                )
+
+            if is_self_application:
+                self.is_volunteer_duplicate = True
+                error_message = _("You have already submitted a request for this pick.")
+            else:
+                error_message = _("This person has already submitted a request for this pick.")
```
2ynn, line 101:
```suggestion
            if  auth_user == self.harvest.pick_leader:
                raise forms.ValidationError(
                    _("A pick leader cannot volunteer for their own harvest.")
                )

            if auth_user == self.request_user:
                self.is_volunteer_duplicate = True
                error_message = _("You have already submitted a request for this pick.")
            else:
                error_message = _("This person has already submitted a request for this pick.")
```

saskatoon/harvest/forms.py
2ynn: why not simply raising a validation error?

saskatoon/harvest/forms.py
```diff
@@ -57,19 +58,48 @@ class Meta:
...
+
+            is_target_harvest_leader = auth_user == self.harvest.pick_leader
+
+            if is_target_harvest_leader:
+                raise forms.ValidationError(
+                    _("The harvest pick leader cannot volunteer for their own pick.")
+                )
+
+            if is_self_application:
+                self.is_volunteer_duplicate = True
```
2ynn, line 98: Please explain why raising a `forms.ValidationError` is not suitable here and why the `if self.request.GET.get('from') == 'calendar':` check in `form_invalid` is necessary

saskatoon/sitebase/templates/app/forms/participation_create_form.html
```diff
@@ -55,17 +55,6 @@ <h2>
...
-        <div class="container">
-            <div class="row">
-                <div class="col-xs-12">
-                    <div class="text-center alert alert-warning" role="alert">
-                        {{ form.non_field_errors }}
-                    </div>
-                </div>
-            </div>
-        </div>
-    {% endif %}
```
2ynn, line 68: Why do we no longer want to display validation errors in the form? 

> redundant error message removed from the request for participation form

If I'm not mistaken only field errors are redundant.? What happens If I want to raise a new general validation error from `def clean()` (i.e. not associated with any particular field but a combination of multitple ones) ? Where will the error be displayed then?

2ynn (changes requested):

saskatoon/harvest/forms.py
JonGlaab: The calendar check is to redirect properly as per the issue #484 instructions. 
If the user is anonymous or not the leader and they came from the calendar, they are redirected to the calendar with the 'You have already submitted a request for this pick.' message.
if the user came from the harvest page, they are redirected back there with the same message. 
if the user is the pick leader trying to volunteer or sign up a duplicate friend, they stay on the form page with the appropriate message (I'm assuming they may want to sign up other people but I could redirect them to the harvest as well)

When redirecting users back to the calendar after they trigger the double application error, there is no form on the calendar page to display the RFPform validation error. I need the 'self.is_volunteer_duplicate = True' tag to have something to pass along with the redirect to display the message correctly on the calendar page.

saskatoon/sitebase/templates/app/forms/participation_create_form.html
JonGlaab: I believe Crispy Forms will render those at the top of the form without the extra block and I thought it was simpler to remove this than block Crispy. 

I added a clean block to raise a non-field error to check myself
selfCheck

saskatoon/harvest/forms.py
JonGlaab: I have the extra logic to ensure anon users are caught and redirected to the calendar as per #484. otherwise they stay on the form

2ynn: Why is it an issue if they stay on the form though?
related to https://github.com/LesFruitsDefendus/saskatoon-ng/pull/783#discussion_r4058369280

2ynn: You set `self.is_volunteer_duplicate = True` when `is_self_application` is True:
```
if is_self_application:
    self.is_volunteer_duplicate = True
```

which means the user requesting *is* authenticated according to:
```
s_self_application = (
                self.request_user
                and self.request_user.is_authenticated
                and self.request_user.email == email
            )
```
so I believe my version is equivalent. That being said I agree it's flawed and I also find the name `is_volunteer_duplicate` confusing. I would read it as "this user is a volunteer (i.e. unauthenticated) and they applied twice)

Basically I think you'd want to set `self.is_duplicate = True` as soon as you find a matching RFP:
```
if (
            target_user
            and target_user.person
            and RFP.objects.filter(person=target_user.person, harvest_id=self.harvest.id).exists()
        ):
        self.is_duplicate = True
        ...
```

saskatoon/harvest/forms.py
2ynn: @JonGlaab Is your issue resolved? If not we need to get to the bottom of this, you shouldn't have to compare emails here

PR #784, opened 2026-09-15
fix: improve harvest form success message

### LINKED ISSUE #627: Bug: harvest form success message
When saving a Harvest form, the success message should congratulate the user for leading a harvest only if:
- the user is the pickleader
- the status went from schedule/ready to success

Image

JonGlaab:
Fixes #627
----------
## Type of change:
- [ ] Bug fix 
----------
## What's Changed:
- the edit harvest form only gives the unique success message to the pick leader updating the status from 'scheduled' or 'ready'
- the harvest status dropdown is updated to function the same
- **note** if a person updates to 'scheduled' or 'ready', then portions some of the produce, there is nothing to stop them from resetting back to 'adopted'. In such a case they can switch from 'adopted' to 'success' bypassing 'scheduled' and 'ready' (but they needed to be in one of those states to portion produce in the first place) 
--------
## Screenshots:
1. pick leader changes status to success:
Screenshot 2026-09-15 at 12 31 48 PM

2. an admin changed the status:
Screenshot 2026-09-15 at 12 42 16 PM

--------
## Affected URLs:
127.0.0.1:8000/harvest/<int>


saskatoon/harvest/views.py
```diff
@@ -254,16 +258,29 @@ def get_success_message(self, cleaned_data) -> StrOrPromise:
                 self.object.save()
                 return ""
 
-            if (pl := self.object.pick_leader) is not None and (person := pl.person) is not None:
+            transitioned_correctly = getattr(self, 'original_status', None) in [
+                Harvest.Status.SCHEDULED,
+                Harvest.Status.READY,
+            ]
```
2ynn: What does this guard against?

saskatoon/harvest/views.py
```diff
@@ -254,16 +258,29 @@ def get_success_message(self, cleaned_data) -> StrOrPromise:
...
+
+            pl = self.object.pick_leader
+            is_user_pickleader = pl is not None and pl == self.request.user
+
+            if (
+                pl is not None
+                and transitioned_correctly
+                and is_user_pickleader
+                and pl.person is not None
+            ):
```
2ynn: Why are you checking `pl is not None` twice? 

```
if ( 
    (pl := self.object.pick_leader) is not None and
    pl == self.request.user and
    (person := pl.person) is not None
): 
    person.get_harvests_as_pickleader(status=Harvest.Status.SUCCEEDED)
```

2ynn (commented):
Please clarify PR title, e.g. `fix: improve nth harvest led congrats message"

saskatoon/harvest/views.py
JonGlaab: Walrus operator is new to me. Seems cool! 
MyPy wouldn't let me pass without the 'pl is not None' line being explicit.
I'll change it to the above

saskatoon/harvest/views.py
```diff
@@ -254,16 +258,26 @@ def get_success_message(self, cleaned_data) -> StrOrPromise:
                 self.object.save()
                 return ""
 
-            if (pl := self.object.pick_leader) is not None and (person := pl.person) is not None:
+            transitioned_correctly = getattr(self, 'original_status', None) in [
```
2ynn: imo using `getattr(self, ..., None)` syntax can backfire. Suppose we delete `form_valid` override, `transitioned_correctly` will silently become always False
```suggestion
            transitioned_correctly = self.original_status in [
```

of if you insist on `checking if `self.original_status` was set then something must be done when you get `None` (e.g. log an error)

saskatoon/harvest/views.py
```diff
@@ -228,6 +228,10 @@ def get_context_data(self, **kwargs):
     def get_form_kwargs(self, *args, **kwargs):
         return super().get_form_kwargs(*args, **kwargs) | {'yields': self.object.yields}
 
+    def form_valid(self, form):
+        self.original_status = self.get_object().status
```
2ynn: Consider storing what you need, e.g.

```suggestion
        self.was_scheduled_or_ready = self.get_object().status in [
            Harvest.Status.SCHEDULED,
            Harvest.Status.READY,
        ]
```

saskatoon/harvest/views.py
```diff
@@ -254,16 +258,26 @@ def get_success_message(self, cleaned_data) -> StrOrPromise:
...
 
-            if (pl := self.object.pick_leader) is not None and (person := pl.person) is not None:
+            transitioned_correctly = getattr(self, 'original_status', None) in [
+                Harvest.Status.SCHEDULED,
+                Harvest.Status.READY,
+            ]
+
+            if (
+                transitioned_correctly
+                and (pl := self.object.pick_leader) is not None
```
2ynn, line 263: similarly here: is there a concrete scenario where `harvest.pick_leader is None` and if so what should we do?

saskatoon/harvest/views.py
```diff
@@ -578,12 +592,38 @@ def harvest_status_change(request, id):
             _("Please complete fruit distribution before marking the harvest as succeeded."),
         )
     else:
+        transitioned_correctly = harvest.status in [
+            Harvest.Status.SCHEDULED,
+            Harvest.Status.READY,
+        ]
```
2ynn: `transitioned_correctly` only makes sense / is only useful if you know the harvest status is `SUCCEEDED` 
```suggestion
```

saskatoon/harvest/views.py
```diff
@@ -578,12 +592,38 @@ def harvest_status_change(request, id):
...
-        messages.success(
-            request,
-            _("Harvest status successfully set to: {}").format(harvest.get_status_display()),
-        )
+
+        pl = harvest.pick_leader
+        if (
+            pl is not None
+            and request_status == Harvest.Status.SUCCEEDED
+            and transitioned_correctly
```
2ynn:
```suggestion
            and request_status == Harvest.Status.SUCCEEDED
            and self.was_scheduled_or_ready
```

saskatoon/harvest/views.py
```diff
@@ -578,12 +592,38 @@ def harvest_status_change(request, id):
...
+                pl.person.get_harvests_as_pickleader(status=Harvest.Status.SUCCEEDED)
+                .filter(start_date__year=tz.now().date().year)
+                .count()
+            )
+            messages.success(
+                request,
+                _(
+                    "You’ve just led your {} fruit harvest this season! "
+                    "Thank you for supporting your community!"
+                ).format(ordinal(season_count)),
```
2ynn, line 612: How does this translate to French?

saskatoon/harvest/views.py
2ynn: BTW. Given that a harvest must have fruit distribution filled and that for fruit distribution to be done it must be marked as `ready`, why don't we simply check if the original status was READY? Are you sure SCHEDULED applies here?

JonGlaab: I was following the #627 issue but I believe you are correct and it should only be ready that matters

saskatoon/harvest/views.py
JonGlaab: applied the above comment so this can be removed

saskatoon/harvest/views.py
JonGlaab: "Vous avez mené votre {} cueillette de fruits de la saison ! Merci de "
"supporter votre communauté ! " was in place already

saskatoon/harvest/views.py
JonGlaab: I added this because of mypy checks. I don't think there is a concrete case but I can add a redirect home if this is somehow true

saskatoon/harvest/views.py
2ynn: What I mean is how does `ordinal(season_count)` behave in French? "Vous avez mené votre 4th cueillete" doesn't work in french. Please add screenshot

2ynn: Nvm you had it already 
screenshot_2026-09-23_23:20:04
Technically should be 2ème

saskatoon/harvest/locale/fr/LC_MESSAGES/django.po
```diff
@@ -1034,152 +1034,156 @@ msgstr ""
...
-#: harvest/views.py:465
+#: harvest/views.py:474
 msgid "You must be a pickleader to delete a fruit distribution entry!"
 msgstr ""
 "Vous devez être chef de cueillette pour supprimer une entrée dans la "
 "distribution de fruits!"
 
 #: harvest/views.py:480
+msgid "Fruit distribution deleted!"
+msgstr "la distribution de fruits effacé!"
```
```suggestion
msgstr "Distribution supprimée!"
```

saskatoon/harvest/views.py
```diff
@@ -578,12 +587,35 @@ def harvest_status_change(request, id):
...
+            )
```
2ynn: I don't think this logic (and translated string ) should be duplicated. Could you please extract a helper, e.g.  in `Harvest` model:
```
def get_congratulations_message(self) -> Optional[str]:
    if self.pick_leader is None or self.status != Status.READY:
        return None
    
      season_count = (
          self.pick_leader.person.get_harvests_as_pickleader(status=Harvest.Status.SUCCEEDED)
              .filter(start_date__year=tz.now().date().year)
              .count()
          )

      return _(
          "You’ve just led your .... 
      )

